PR #1, opened 2020-09-05
Translation of BIR to LLVM IR

shubhamnarlawar:
Purpose of BIRReader class - 

a. Reads Constant Pool entries.
b. Reads Module consisting of functions, basic blocks and instructions.

BIRPackage class consist of translate() function which translates functions, basic blocks and instructions to the LLVM IR.


shubhamnarlawar:
Please review and suggest changes.

PR #93, opened 2021-02-11
Add support for the windows build

### LINKED ISSUE #66: Make nballerinacc compilable on Windows OS
**Description:**
Add support to build in Windows

**Suggested Labels:**
Type/Improvement,Priority/High 

**Suggested Assignees:**
@KavinduZoysa

KavinduZoysa:
## Purpose
$subject
Resolves #66.

In this PR we only expect to build nballerina executable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,11 @@
 cmake_minimum_required(VERSION 3.2)
+if (NOT UNIX)
+    cmake_policy(SET CMP0091 NEW)
+endif()
 project(nballerina VERSION 0.0.1 LANGUAGES CXX)
 
 # Find and loacd LLVM components
-find_package(LLVM 11 REQUIRED CONFIG)
+find_package(LLVM 11.1.0 REQUIRED CONFIG)
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
 add_definitions(${LLVM_DEFINITIONS})
 
@@ -18,8 +21,23 @@ file(GLOB SOURCES
 ${PROJECT_SOURCE_DIR}/compiler/*.cpp
 )
 add_executable(nballerinacc ${SOURCES})
+
 add_dependencies(nballerinacc BallerinaRT)
-target_link_libraries(nballerinacc PRIVATE LLVM-11)
+
+if (UNIX) 
+    target_link_libraries(nballerinacc PRIVATE LLVM-11)
+else ()
+    set_property(TARGET nballerinacc PROPERTY
+                MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+    target_link_libraries(nballerinacc PRIVATE LLVM-C
+    LLVMBinaryFormat
+    LLVMBitReader
+    LLVMBitstreamReader
+    LLVMCodeGen
+    LLVMCore
+    LLVMRemarks
+    LLVMSupport)
+endif()
 
 # Set extra warnings for nballerinacc
 set(DEBUG_OPTIONS -Wall)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,11 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.15)
 if (NOT UNIX)
     cmake_policy(SET CMP0091 NEW)
 endif()
 project(nballerina VERSION 0.0.1 LANGUAGES CXX)
 
 # Find and loacd LLVM components
-find_package(LLVM 11.1.0 REQUIRED CONFIG)
+find_package(LLVM 11 REQUIRED CONFIG)
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
 add_definitions(${LLVM_DEFINITIONS})
 
@@ -24,24 +24,22 @@ add_executable(nballerinacc ${SOURCES})
 
 add_dependencies(nballerinacc BallerinaRT)
 
-if (UNIX) 
-    target_link_libraries(nballerinacc PRIVATE LLVM-11)
-else ()
+llvm_map_components_to_libnames(llvm_libs codegen)
+if (NOT UNIX)
     set_property(TARGET nballerinacc PROPERTY
                 MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
-    target_link_libraries(nballerinacc PRIVATE LLVM-C
-    LLVMBinaryFormat
-    LLVMBitReader
-    LLVMBitstreamReader
-    LLVMCodeGen
-    LLVMCore
-    LLVMRemarks
-    LLVMSupport)
 endif()
+target_link_libraries(nballerinacc PRIVATE ${llvm_libs})
 
 # Set extra warnings for nballerinacc
-set(DEBUG_OPTIONS -Wall)
-set(RELEASE_OPTIONS -Wall)
+if (MSVC)
+    set(DEBUG_OPTIONS /W4)
+    set(RELEASE_OPTIONS /W4)
+else()
+    set(DEBUG_OPTIONS -Wall)
+    set(RELEASE_OPTIONS -Wall)
+endif()
+
 target_compile_options(nballerinacc PUBLIC "$<$<CONFIG:DEBUG>:${DEBUG_OPTIONS}>")
 target_compile_options(nballerinacc PUBLIC "$<$<CONFIG:RELEASE>:${RELEASE_OPTIONS}>")
 
--- a/README.md
+++ b/README.md
@@ -45,22 +45,24 @@ This will build:
 * Install [cargo](https://doc.rust-lang.org/cargo/getting-started/installation.html) and run `cargo install cbindgen`
 * Install [python-3](https://www.python.org/downloads/)
 * Build LLVM from source
-  * Install Visual Studio 2019. Please make sure that **Desktop development with C++** should be installed with **C++ CMake tools for Windows** and **C++ ATL for latest v142 build tools (x86 & x64)** which are optional.
-  * Download the LLVM source from [here](https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.1/llvm-11.0.1.src.tar.xz) and extract. (Let us consider the extracted path as `LLVM_SRC_DIR_PATH`)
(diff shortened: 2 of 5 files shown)


ruvi-d (changes requested):
I think we should include the LLVM src as a submodule as discussed in #53. We only need to init and build it under Windows. 

Doing this will make our CMake script simpler because we know the relative path to LLVM and it will also make it easier to setup CI/CD for windows.

compiler/BIRReader.cpp
```diff
@@ -143,7 +148,7 @@ TypeDecl *ConstantPoolSet::getTypeCp(uint32_t index, bool voidToInt) {
   if (typeDecl->getTypeDeclName() == "") {
     char newName[20];
     char *p;
-    p = stpcpy(newName, "anon-");
+    p = strcpy(newName, "anon-");
```
ruvi-d, line 151: You can't replace `stpcpy` with `strcpy`. One returns a pointer to the start, the other returns a pointer to the end.

compiler/llvmTypeCastInsn.cpp
```diff
@@ -1,4 +1,7 @@
 #include "BIR.h"
+#ifndef unix 
+    #define __attribute__(unused)
+#endif
```
ruvi-d, line 4: As mentioned previously. Let's use `volatile` instead.

README.md
```diff
@@ -28,3 +30,29 @@ This will build the Rust runtime (output in runtime/target/release), the C++ hea
...
+### Prerequisites
+* Install [cargo](https://doc.rust-lang.org/cargo/getting-started/installation.html) and run `cargo install cbindgen`
+* Install [python-3](https://www.python.org/downloads/)
+* Build LLVM from source
+  * Install Visual Studio 2019. Please make sure that **Desktop development with C++** should be installed with **C++ CMake tools for Windows** and **C++ ATL for latest v142 build tools (x86 & x64)** which are optional.
+  * Download the LLVM source from [here](https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.1/llvm-11.0.1.src.tar.xz) and extract. (Let us consider the extracted path as `LLVM_SRC_DIR_PATH`)
+  * Create the `build` directory. Run following commands in **x64 Native Tools Command Prompt for VS 2019**.
+    
+        cd build
+        cmake -DCMAKE_BUILD_TYPE="Release" -Thost=x64 -G "Visual Studio 16 2019" -A x64 -DLLVM_USE_CRT_RELEASE=MT ../
```
ruvi-d, line 45: Better to set the `-DCMAKE_INSTALL_PREFIX="<path>"` flag too

README.md
```diff
@@ -28,3 +30,29 @@ This will build the Rust runtime (output in runtime/target/release), the C++ hea
...
         make check
 * To add LIT tests, add LIT compatible .bal files to the `test/` folder
+
+## Building from source in Windows 10(x64)
+
+### Prerequisites
+* Install [cargo](https://doc.rust-lang.org/cargo/getting-started/installation.html) and run `cargo install cbindgen`
+* Install [python-3](https://www.python.org/downloads/)
+* Build LLVM from source
+  * Install Visual Studio 2019. Please make sure that **Desktop development with C++** should be installed with **C++ CMake tools for Windows** and **C++ ATL for latest v142 build tools (x86 & x64)** which are optional.
```
ruvi-d, line 40: No need to install Visual Studio 2019 IDE. The build tools alone is sufficient since we are build on the command line. https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019

ruvi-d (changes requested):
Noted some issues and also some suggestions for improvements.

CMakeLists.txt
```diff
@@ -19,7 +22,21 @@ ${PROJECT_SOURCE_DIR}/compiler/*.cpp
...
+                MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+    target_link_libraries(nballerinacc PRIVATE LLVM-C
+    LLVMBinaryFormat
+    LLVMBitReader
+    LLVMBitstreamReader
+    LLVMCodeGen
+    LLVMCore
+    LLVMRemarks
+    LLVMSupport)
+endif()
```
ruvi-d, line 39: We can make this better as :
```
llvm_map_components_to_libnames(llvm_libs codegen)
if (NOT UNIX)
    set_property(TARGET nballerinacc PROPERTY
                MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
endif()
target_link_libraries(codegen_test PRIVATE ${llvm_libs})
```

ruvi-d (changes requested):
Suggested change to the CMake's Windows logic

CMakeLists.txt
ruvi-d: ^ this will then statically link LLVM libs for both Unix and Windows targets.

CMakeLists.txt
```diff
@@ -28,4 +45,4 @@ target_compile_options(nballerinacc PUBLIC "$<$<CONFIG:DEBUG>:${DEBUG_OPTIONS}>"
 target_compile_options(nballerinacc PUBLIC "$<$<CONFIG:RELEASE>:${RELEASE_OPTIONS}>")
```
ruvi-d, line 45: `-Wall` is not valid for MSVC. Need to check and add compiler specific flags as shown here: https://alandefreitas.github.io/moderncpp/cmake-functions/target-options/

KavinduZoysa: Changed

CMakeLists.txt
KavinduZoysa: Changed

README.md
KavinduZoysa: Fixed

README.md
KavinduZoysa: Mentioned that.

compiler/BIRReader.cpp
```diff
@@ -1,6 +1,11 @@
 #include "BIRReader.h"
 #include "BIR.h"
-#include <libgen.h>
+#ifdef unix 
+    #include <libgen.h> 
+#else 
+    #define __attribute__(unused)
+    #define random() rand()
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed

compiler/BIRReader.cpp
KavinduZoysa: @ruvi-d could you please suggest an option here?

compiler/BIRReader.cpp
```diff
@@ -1,6 +1,11 @@
 #include "BIRReader.h"
 #include "BIR.h"
-#include <libgen.h>
+#ifdef unix 
+    #include <libgen.h> 
+#else 
+    #define __attribute__(unused)
```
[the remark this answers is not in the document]
KavinduZoysa: Could you please explain the relationship between `__attribute__(unused)` and `volatile`? 

Are you suggesting to use `volatile`, to turn off the compiler optimizations for a particular variable?

CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
 cmake_minimum_required(VERSION 3.2)
+if (NOT UNIX)
+    cmake_policy(SET CMP0091 NEW)
+endif()
```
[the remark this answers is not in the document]
KavinduZoysa: Are you suggesting here to change the version?

compiler/BIRReader.cpp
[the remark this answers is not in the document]
ayonam: Use of volatile is not a good idea.  The attribute is the better one as it states the true nature of the variable, i.e., unused.  Volatile indicates that there is a use of this variable that may or may not be obvious to the compiler from the plain parsing of the code.  While this attribute tells that the variable is indeed unused.  It is just that Windoze compilers don't understand what most industry standard compilers do; they have to be different in order to assert their technological relevance in the face of obscurity.  So we define this attribute as an empty string on Windoze.

CMakeLists.txt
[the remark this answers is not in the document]
ruvi-d: yes :-)

compiler/BIRReader.cpp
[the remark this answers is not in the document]
ruvi-d: @ayonam but if we define attribute as empty string, wouldn't the MSVC compiler optimize away the "dead load" (because it's unused from the compiler perspective). In the past I have have used the volatile keyword to prevent these kinds of "dead loads" and "dead stores" getting optimized away when accessing controller chip registers.
Correct me if otherwise, but isn't the `__attribute__(unused)` also serving the same purpose here? Preventing the "dead load" of reading the next bytes of the BIR, from getting optimized away?

README.md
ruvi-d: Let make it a MUST to include this flag. A folder called `install` in the llvm-src dir might be good. Otherwise it looks like it installs to the include folder in your case. That would be a bad idea.

compiler/BIRReader.cpp
ruvi-d: is `stpcpy` not available in MSVCP? If you must use `strcpy` I suppose you could add the length of the "anon-" literal to 'p' to get the same pointer as  `stpcpy`

compiler/BIRReader.cpp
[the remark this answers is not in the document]
ayonam: @ruvi-d Actually, here we only want to suppress the warning.  What you did in the embedded domain, is the standard practice there because there are situations in an embedded processor where a register may be required to be read without the data being used for any purpose.  E.g., reading a piece of data from an I/O Port, where the data being read is not of importance but the succeeding data might be of importance.  But the I/O ports being serial in nature, you cannot read the next data until you have read and discarded the current one.  So in order to force the read operation, the variables are declared as volatile.  Here, our read operations will happen regardless since they are read using a function call.  What we need is a suppression of the warnings and that is best done with the attribute.  This attribute does not control the optimization.

README.md
KavinduZoysa: @Ruvi, I have set the `-DCMAKE_INSTALL_PREFIX="../install"`. But my `install` folder is empty. If I am wrong could you please provide me the correct way to mention the path with folder structure?

CMakeLists.txt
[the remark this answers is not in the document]
KavinduZoysa: Fixed.

README.md
KavinduZoysa: Solved it and updated.

uthaiyashankar:
@KavinduZoysa There are some conflicts due to recent PR merge. Can you fix them too please?

compiler/BIRReader.cpp
KavinduZoysa: Fixed.

KavinduZoysa:
> @KavinduZoysa There are some conflicts due to recent PR merge. Can you fix them too please?

Fixed.

uthaiyashankar (approved):

PR #102, opened 2021-02-17
Implement map init and store for Int types

### LINKED ISSUE #89: Implement map literal and access
**Description:**
```
map<int> marks = {sam: 50, jon: 60};
any m = marks["sam"]; // assert 50
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

ruvi-d:
## Purpose
Addresses two sub tasks (Map init and store for Int member types only) listed under #89 
closes #89


ruvi-d:
Note that there is currently no codegen logic to deinit the Map object. Suggestions are welcome.

compiler/llvmFunctionGen.cpp
```diff
@@ -170,20 +171,20 @@ LLVMTypeRef BIRFunction::getLLVMTypeRefOfType(TypeDecl *typeD) {
     return LLVMPointerType(LLVMInt8Type(), 0);
   case TYPE_TAG_ANY:
     return LLVMPointerType(LLVMInt64Type(), 0);
+  case TYPE_TAG_MAP:
+    return LLVMPointerType(LLVMInt64Type(), 0);
```
KavinduZoysa, line 175: Shall we combine these 2 cases?

ruvi-d: Will do

compiler/llvmFunctionGen.cpp
```diff
@@ -169,21 +170,20 @@ LLVMTypeRef BIRFunction::getLLVMTypeRefOfType(TypeDecl *typeD) {
   case TYPE_TAG_STRING:
     return LLVMPointerType(LLVMInt8Type(), 0);
   case TYPE_TAG_ANY:
+  case TYPE_TAG_MAP:
     return LLVMPointerType(LLVMInt64Type(), 0);
```
KavinduZoysa, line 174: @ruvi-d , just to know, what is the reason to set the i64* for map type? What is the theory behind that?

ruvi-d: I didn't have a good reason to pick that. Ideally, I would have liked to use something akin to a `void *` so that there is some protection against accidental de-referencing.

KavinduZoysa: Thanks @ruvi-d

shubhamnarlawar:
@ruvi-d - This PR looks fine to me!

shubhamnarlawar:
@ayonam - Can you please take quick look at this PR to give any suggestions if required?

ramakotareddy:
@ruvi-d  i have gone through Translation side code, Code looks good to me.
@ayonam Can you please give your comments.

ruvi-d:
@shubhamnarlawar and @ramakotareddy noted thanks. Please use the Github option "Review changes" and then select "Approve / Request changes" to complete your review

shubhamnarlawar (approved):
Good to merge!

KavinduZoysa:
`map<int> marks = {}`
@ruvi-d , I got the .ll file for above simple code and it is as followed.

```
define void @main() {
entry:
  %"%0" = alloca i32, align 4
  %"%1" = alloca i64*, align 8
  %"%2" = alloca i32, align 4
  br label %bb0

bb0:                                              ; preds = %entry
  %0 = call i64* @map_new_int()
  store i64* %0, i64** %"%1", align 8
  br label %bb2

bb1:                                              ; No predecessors!
  br label %bb2

bb2:                                              ; preds = %bb1, %bb0
  ret void
}
```

Here we assign the returned value to `%0`. I think it will be more clear if we can assign it to temp value like %1_temp and store it in %1, WDYT?

ruvi-d:
> Here we assign the returned value to `%0`. I think it will be more clear if we can assign it to temp value like %1_temp and store it in %1, WDYT?

I am a bit torn about that. You are right that it would improve readability, but it is a redundant step that would likely get optimized out when we add LLVM IR optimization passes later on. 

@ayonam : any thoughts?

KavinduZoysa:
> > Here we assign the returned value to `%0`. I think it will be more clear if we can assign it to temp value like %1_temp and store it in %1, WDYT?
> 
> I am a bit torn about that. You are right that it would improve readability, but it is a redundant step that would likely get optimized out when we add LLVM IR optimization passes later on.
> 
> @ayonam : any thoughts?

Yes, redundancy point of view, I agree with you.

KavinduZoysa (approved):

ruvi-d:
Summary of open question to @ayonam 

1. There is currently no codegen logic to deinit (free) the Map object. Is there a recommended approach to call deinit for the local variables at least?
2. To hold the map pointer returned from Rust side, is there a preference between using *i8 or *i64 on the codegen side? 
3. When we are generating LLVM IR code, is it better to use more temp variables and be verbose? or is it better to use as few as possible and assign directly to local variables?

compiler/llvmMap.cpp
```diff
@@ -0,0 +1,65 @@
...
+  BIRFunction *funcObj = getFunction();
+  BIRPackage *pkgObj = getPkgAddress();
+  string lhsName = getLhsOperand()->name();
+  string rhsName = rhsOp->name();
+  LLVMBuilderRef builder = funcObj->getLLVMBuilder();
+
+  // Find VarDecl corresponding to lhs to determine member type
+  VarDecl *lhsVar = funcObj->getNameVarDecl(lhsName);
+  assert(lhsVar);
+  assert(lhsVar->getVarKind() == LOCAL_VAR_KIND);
```
ayonam, line 19: Why is this assertion?  Can't the LHS var be a global variable?

compiler/llvmMap.cpp
```diff
@@ -0,0 +1,65 @@
...
+    assert(false);
+  }
+
+  // Codegen for Map of Int type store
+  LLVMValueRef mapStoreFunc =
+      pkgObj->getFunctionRefBasedOnName("map_store_int");
+  if (!mapStoreFunc)
+    mapStoreFunc = getMapIntStoreDeclaration(modRef, pkgObj);
+
+  LLVMValueRef lhsOpTempRef = funcObj->getLocalToTempVar(getLhsOperand());
```
ayonam, line 37: I think you should check the scope of LHS operand and then either call this getLocalToTempVar() if it is a local variable or else do something like a GEP (Global Element Pointer).

compiler/llvmMap.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  LLVMValueRef rhsOpRef = funcObj->getLocalVarRefUsingId(rhsName);
+  LLVMValueRef keyRef = funcObj->getLocalToTempVar(keyOp);
```
ayonam, line 40: Ditto for the RHS op and the KEY op.

compiler/llvmMap.cpp
```diff
@@ -0,0 +1,65 @@
...
+  LLVMTypeRef int32PtrType = LLVMPointerType(LLVMInt32Type(), 0);
+  LLVMTypeRef charArrayPtrType = LLVMPointerType(LLVMInt8Type(), 0);
+  LLVMTypeRef memPtrType = LLVMPointerType(LLVMInt64Type(), 0);
+  paramTypes[0] = memPtrType;
+  paramTypes[1] = charArrayPtrType;
+  paramTypes[2] = int32PtrType;
+  LLVMTypeRef funcType = LLVMFunctionType(LLVMVoidType(), paramTypes, 3, 0);
+  LLVMValueRef addedFuncRef =
+      LLVMAddFunction(modRef, "map_store_int", funcType);
+  pkg->addArrayFunctionRef("map_store_int", addedFuncRef);
```
ayonam, line 63: This is fine for now.  But going forward, as we add different kinds of Maps with different data types, should we be having polymorphic functions on the Rust side or do we have different names for each data type?  If it is the former, then what should this function return?  Probably it should return nothing.  It should simply setup the types of each of the polymorphic functions and return void.

compiler/llvmStructure.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+void StructureInsn::mapInsnTranslate(VarDecl *lhsVar, LLVMModuleRef &modRef) {
+
+  BIRFunction *funcObj = getFunction();
+  BIRPackage *pkgObj = getPkgAddress();
+  LLVMBuilderRef builder = funcObj->getLLVMBuilder();
+  string lhsName = getLhsOperand()->name();
+  LLVMValueRef lhsOpRef = funcObj->getLocalVarRefUsingId(lhsName);
+  if (!lhsOpRef)
+    lhsOpRef = pkgObj->getGlobalVarRefUsingId(lhsName);
```
ayonam, line 33: Here you are checking for both local as well as global var refs.  You need to do like this everywhere.

compiler/llvmStructure.cpp
```diff
@@ -0,0 +1,65 @@
...
+  assert(lhsVar->getTypeDecl()->getTypeTag() == TYPE_TAG_MAP);
+  MapTypeDecl *mapTypeDelare =
+      static_cast<MapTypeDecl *>(lhsVar->getTypeDecl());
+
+  // Get member type
+  int memberTypeTag = mapTypeDelare->getTypeMemberTag();
+  // Only handle Int type
+  if (memberTypeTag != TYPE_TAG_INT) {
+    std::cerr << "Non INT type maps are currently not supported" << std::endl;
+    assert(false);
```
ayonam, line 44: You should call llvm_unreachable() instead of assert(false)

compiler/llvmStructure.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+// Declaration for new map<int> function
+LLVMValueRef StructureInsn::getNewMapIntDeclaration(LLVMModuleRef &modRef,
+                                                    BIRPackage *pkg) {
+  LLVMTypeRef memPtrType = LLVMPointerType(LLVMInt64Type(), 0);
+  LLVMTypeRef funcType = LLVMFunctionType(memPtrType, nullptr, 0, 0);
+  LLVMValueRef addedFuncRef = LLVMAddFunction(modRef, "map_new_int", funcType);
+  pkg->addArrayFunctionRef("map_new_int", addedFuncRef);
+  return addedFuncRef;
+}
```
ayonam, line 65: Again the same comment here regarding polymorphic functions on the Rust side to handle initialization of maps of different data types.

test/map/mapBasic.bal
```diff
@@ -0,0 +1,7 @@
+// RUN: JAVA_HOME=%java_path %testRunScript %s %nballerinacc | filecheck %s
+int _bal_result = 0;
+public function main() {
+    map<int> marks = {};
+    marks["jon"] = 60;
+}
+// CHECK: length=1
```
ayonam, line 7: Shouldn't there also be a check to see if the marks["jon"] is indeed 60 or not?

ayonam (changes requested):

compiler/BIR.h
```diff
@@ -65,7 +65,7 @@ enum InstructionKind {
   INSTRUCTION_KIND_MOVE = 20,
   INSTRUCTION_KIND_CONST_LOAD = 21,
   INSTRUCTION_KIND_NEW_STRUCTURE,
-  INSTRUCTION_KIND_NEW_MAP,
```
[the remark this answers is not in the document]
ruvi-d: There is no "new map" instruction as per the BIR spec listed here: https://github.com/ballerina-platform/ballerina-lang/blob/bd8e12ac24a70079247d1b178de888cb0e76a96a/docs/bir-spec/src/main/resources/kaitai/bir.ksy#L988

We get `new struct` instruction with a LHS value of type Map for new Maps

compiler/BIRReader.cpp
```diff
@@ -139,23 +140,35 @@ bool ConstantPoolSet::getBooleanCp(uint32_t index) {
...
+    return new TypeDecl(TYPE_TAG_INT, name, shapeCp->getTypeFlag());
+
+  // Handle Map type
+  if (type == TYPE_TAG_MAP) {
+    ConstantPoolEntry *shapeEntry =
+        getEntry(shapeCp->getConstraintTypeCpIndex());
+    assert(shapeEntry->getTag() ==
+           ConstantPoolEntry::tagEnum::TAG_ENUM_CP_ENTRY_SHAPE);
+    ShapeCpInfo *typeShapeCp = static_cast<ShapeCpInfo *>(shapeEntry);
+    int memeberType = typeShapeCp->getTypeTag();
```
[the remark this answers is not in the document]
ruvi-d: Thanks for spotting it. Will fix.

compiler/llvmFunctionGen.cpp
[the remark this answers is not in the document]
ruvi-d: Please see this question regarding this topic: https://github.com/ballerina-platform/nballerina/pull/102#issuecomment-785677115
We were not sure whether it is best to use a int-8 or int-64 pointer to hold the address returned from the Rust runtime.

compiler/BIR.h
```diff
@@ -754,8 +795,7 @@ class BIRFunction : public BIRNode {
   Operand *getParam(int i) { return requiredParams[i]; }
   VarDecl *getReceiver() { return receiver; }
   Param *getRestParam() { return restParam; }
-  vector<VarDecl *> getLocalVars() { return localVars; }
-  VarDecl *getLocalVar(int i) { return localVars[i]; }
```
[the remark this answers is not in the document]
ruvi-d: We are not using them anymore

compiler/BIR.h
```diff
@@ -544,12 +582,15 @@ class TypeDescInsn : public NonTerminatorInsn {
 class StructureInsn : public NonTerminatorInsn {
 private:
   Operand *rhsOp;
+  void mapInsnTranslate(VarDecl *lhsVar, LLVMModuleRef &modRef);
+  LLVMValueRef getNewMapIntDeclaration(LLVMModuleRef &modRef, BIRPackage *pkg);
```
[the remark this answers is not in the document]
ruvi-d: The BIR for new Map is actually a `new structure` with a LHS of type Map. So I am having to handle the new map Code gen from the StrutureIsn class. Hope that makes sense?

compiler/BIR.h
```diff
@@ -777,9 +817,10 @@ class BIRFunction : public BIRNode {
   void setParam(Operand *param) { requiredParams.push_back(param); }
   void setReceiver(VarDecl *var) { receiver = var; }
   void setRestParam(Param *param) { restParam = param; }
-  void setLocalVars(vector<VarDecl *> l) { localVars = l; }
   void setNumParams(int paramcount) { paramCount = paramcount; }
-  void setLocalVar(VarDecl *var) { localVars.push_back(var); }
+  void setLocalVar(string name, VarDecl *var) {
+    localVars.insert(pair<string, VarDecl *>(name, var));
+  }
```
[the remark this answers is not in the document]
ruvi-d: Yes makes sense. Fill fix.

compiler/llvmMap.cpp
```diff
@@ -0,0 +1,65 @@
+#include "BIR.h"
+
+// new Map Instruction and Codegen logic are in the llvmStructure.cpp
```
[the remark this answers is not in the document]
ruvi-d: There is no `new map` BIR instruction. What we are getting is a `new structure` instruction withe the RHS that has a Map type. That is why I am handling the codegen for new map as part of the new structure instruction class

compiler/llvmMap.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+MapStoreInsn::MapStoreInsn(Location *pos, InstructionKind kind, Operand *lOp,
+                           Operand *KOp, Operand *rOp)
+    : NonTerminatorInsn(pos, kind, lOp), keyOp(KOp), rhsOp(rOp) {}
+
```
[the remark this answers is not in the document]
ruvi-d: You are only supposed to add inline method definitions to the header right?

compiler/llvmMap.cpp
ruvi-d: will check for global variables and get back.

compiler/llvmMap.cpp
ruvi-d: noted. will check with global variables

compiler/llvmMap.cpp
ruvi-d: noted. thanks

compiler/llvmStructure.cpp
ruvi-d: noted. thanks

test/map/mapBasic.bal
ruvi-d: That will be handled as par of #128

compiler/llvmFunctionGen.cpp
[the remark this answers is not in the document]
ruvi-d: I've pushed a change to use int8* instead of int64*

ruvi-d:
Addressed the review comments from Ayonam. The discussion about why the new map instruction is handled instead the new Structure instruction is unresolved.

ayonam:
@ruvi-d   @KavinduZoysa 

Please find my replies inlined below:

> Summary of open question to @ayonam
> 
>     1. There is currently no codegen logic to deinit (free) the Map object. Is there a recommended approach to call deinit for the local variables at least?
> 

Local variables that are created as part of LLVM generated code will automatically get deallocated once the current function returns and we need not do anything about it.  For those that are created by the Rust side of things through an explicit allocation on the heap, we need to deallocate them explicitly (or through a memory manager / garbage collector).  So although these are local variables from the programmer's perspective, they are actually allocated by Rust on the heap.  

>     2. To hold the map pointer returned from Rust side, is there a preference between using *i8 or *i64 on the codegen side?

Actually since we treat that pointer as an opaque pointer, it is equivalent to void* in the C/C++ paradigm.  Now in LLVM there is no concept of a void pointer.  A void pointer is represented as an i8*.  So I think we should treat such map pointers as i8*.

> 
>     3. When we are generating LLVM IR code, is it better to use more temp variables and be verbose? or is it better to use as few as possible and assign directly to local variables?

Don't bother about the temp variables.  The optimizer downstream is going to straighten them out.  Currently we are not turning on the optimizer (by passing -O0) and hence you still see all those temps in the LLVM dump.  We are doing so, so that we can debug things more easily.  Once we reach certain level of maturity in our codegen, we will turn on the optimizer and then you should see most of these temps going away.

ruvi-d:
Noted.. Thanks for the answers Ayonam

KavinduZoysa (approved):

uthaiyashankar (approved):

ayonam:
> Addressed the review comments from Ayonam. The discussion about why the new map instruction is handled instead the new Structure instruction is unresolved.

@Ruvi, Do we have a resolution to this one as yet?  If yes, then this should be good to approve.  If not, then I'd prefer if you can open a separate Issue about this and we can discuss it there.

ruvi-d:
> > Addressed the review comments from Ayonam. The discussion about why the new map instruction is handled instead the new Structure instruction is unresolved.
> 
> @ruvi, Do we have a resolution to this one as yet? If yes, then this should be good to approve. If not, then I'd prefer if you can open a separate Issue about this and we can discuss it there.

No resolution yet. I've started a slack thread on it too here: https://ballerina-platform.slack.com/archives/C01LNSAHK18/p1614654420035700
Let's close this PR and I will create an Github issue to track it separately.

ayonam (approved):

PR #105, opened 2021-02-17
Add support for Array Type Checker

### LINKED ISSUE #86: Implement is_same_type for array type
**Description:**
Implement `is_same_type()` for array type in RUST.

Input - one source type mangled string and one destination type mangled string.
Output - `bool` value stating whether typecast is possible or not.

More details are available in discussion - #50 

**Suggested Assignees:**
@shubhamnarlawar

shubhamnarlawar:
## Purpose
Checks whether array typecast is possible from source type to destination type.

Related discussion - https://github.com/ballerina-platform/nballerina/discussions/50

Fixes #86


shubhamnarlawar:
Please review and suggest changes - @ruvi-d @manuranga

shubhamnarlawar:
This PR addresses issue #86.

shubhamnarlawar:
From issue - #86 

`Write wrapper function for is_same_type() to check whether original types of source and destination can be type casted.`

The original types are available in smart pointers. The original types - for example - int, char, any, etc is supposed to be checked before callling `is_same_type()`.

As discussed earlier, wrapper function is supposed to be implemented in rust. `A` represents array in `is_same_type()`. Probably I should take `a` for any data type in wrapper function.

Is wrapper function suggested to be implemented in rust or llvm side?

ruvi-d:
> From issue - #86
> 
> `Write wrapper function for is_same_type() to check whether original types of source and destination can be type casted.`
> 
> The original types are available in smart pointers. The original types - for example - int, char, any, etc is supposed to be checked before callling `is_same_type()`.
> 
> As discussed earlier, wrapper function is supposed to be implemented in rust. `A` represents array in `is_same_type()`. Probably I should take `a` for any data type in wrapper function.
> 
> Is wrapper function suggested to be implemented in rust or llvm side?

IMO, such decisions should be based on whether the check is at compile time or runtime. If the check needs to happen at compile time, the implementation should be on the C++/LLVM side. If it's a runtime check, then it should happen on the Rust side.

manuranga:
Hi @shubhamnarlawar ,
Is it possible to directly call the rust function from generated code, without using a wrapper ?

runtime/src/lib.rs
```diff
@@ -19,84 +19,143 @@
...
+    if source == destination {
+        return true;
+    }
+    //Index 0 represents type of data structure
+    if source.chars().nth(0) != destination.chars().nth(0) {
+        return false;
+    }
+    let mut src_priority = 0;
+    let mut dest_priority = 0;
+    match source.chars().nth(0) {
```
ruvi-d, line 79: Define these magic numbers for bit positions as constants.

runtime/src/lib.rs
```diff
@@ -19,84 +19,143 @@
...
 use std::mem;
 use std::os::raw::c_char;
 
-// String comparision
+pub fn compute_size(type_string: &str) -> i32 {
+    let mut bits = String::new();
+    //Index 2 and 3 represent the size
+    bits.push(type_string.chars().nth(2).unwrap());
+    bits.push(type_string.chars().nth(3).unwrap());
+    let total_bits: i32 = bits.parse().unwrap();
```
ruvi-d, line 31: Need to handle unbounded Array sizes

ruvi-d (changes requested):
Please see review comments

ruvi-d:
@shubhamnarlawar other general review comments:

1. Shall we move the logic out of lib.rs and into separate modules? Let's try to keep lib.rs as thin as possible

2. Add some Rust unit tests for the new functions

3. Do we need to handle unbounded size arrays? I believe that on the BIR side unbounded is handled with a `-1` size. How are we handling that?

shubhamnarlawar:
> > From issue - #86
> > `Write wrapper function for is_same_type() to check whether original types of source and destination can be type casted.`
> > The original types are available in smart pointers. The original types - for example - int, char, any, etc is supposed to be checked before callling `is_same_type()`.
> > As discussed earlier, wrapper function is supposed to be implemented in rust. `A` represents array in `is_same_type()`. Probably I should take `a` for any data type in wrapper function.
> > Is wrapper function suggested to be implemented in rust or llvm side?
> 
> IMO, such decisions should be based on whether the check is at compile time or runtime. If the check needs to happen at compile time, the implementation should be on the C++/LLVM side. If it's a runtime check, then it should happen on the Rust side.

Got it. Thanks for the clarification.

shubhamnarlawar:
> @shubhamnarlawar other general review comments:
> 
>     1. Shall we move the logic out of lib.rs and into separate modules? Let's try to keep lib.rs as thin as possible
> 
>     2. Add some Rust unit tests for the new functions
> 
>     3. Do we need to handle unbounded size arrays? I believe that on the BIR side unbounded is handled with a `-1` size. How are we handling that?

1. Yes. How about a new file `typeChecker.rs`?
2. Sure. I will add unit tests at last in this PR itself after pushing above review changes.
3. Let me check and get back to you on unbounded size arrays.

runtime/src/lib.rs
shubhamnarlawar: Can you elaborate a bit more in this suggestion?

Do you mean to use below snippet instead -

 `   let constant = source.chars().nth(0);`
`    match constant {`

ruvi-d:
> > > From issue - #86
> > > `Write wrapper function for is_same_type() to check whether original types of source and destination can be type casted.`
> > > The original types are available in smart pointers. The original types - for example - int, char, any, etc is supposed to be checked before callling `is_same_type()`.
> > > As discussed earlier, wrapper function is supposed to be implemented in rust. `A` represents array in `is_same_type()`. Probably I should take `a` for any data type in wrapper function.
> > > Is wrapper function suggested to be implemented in rust or llvm side?
> > 
> > 
> > IMO, such decisions should be based on whether the check is at compile time or runtime. If the check needs to happen at compile time, the implementation should be on the C++/LLVM side. If it's a runtime check, then it should happen on the Rust side.
> 
> Got it. Thanks for the clarification.

Let us know if there is a good reason to use Rust logic for compile time checks and then maybe we can take it case by case.

ruvi-d:
> 1. Yes. How about a new file `typeChecker.rs`?

I think it should be `type_checker.rs` inline with Rust naming conversions.

runtime/src/lib.rs
ruvi-d: no.. I mean define the string indexes as constants like:
```
const BASE_TYPE_INDX: u32 = 0;
const ARRAY_MEMBER_TYPE_INDX: u32 = 1;
const ARRAY_SIZE_LSB_INDX: u32 = 2;
const ARRAY_SIZE_MSB_INDX: u32 = 3;
...
// Then use these constants when accessing the chars
match source.chars().nth(BASE_TYPE_INDX)
```
Hope that makes sense?

shubhamnarlawar:
> > > > From issue - #86
> > > > `Write wrapper function for is_same_type() to check whether original types of source and destination can be type casted.`
> > > > The original types are available in smart pointers. The original types - for example - int, char, any, etc is supposed to be checked before callling `is_same_type()`.
> > > > As discussed earlier, wrapper function is supposed to be implemented in rust. `A` represents array in `is_same_type()`. Probably I should take `a` for any data type in wrapper function.
> > > > Is wrapper function suggested to be implemented in rust or llvm side?
> > > 
> > > 
> > > IMO, such decisions should be based on whether the check is at compile time or runtime. If the check needs to happen at compile time, the implementation should be on the C++/LLVM side. If it's a runtime check, then it should happen on the Rust side.
> > 
> > 
> > Got it. Thanks for the clarification.
> 
> Let us know if there is a good reason to use Rust logic for compile time checks and then maybe we can take it case by case.

Sure.

 At this point, there is no need to write wrapper function for calling 'is_same_type()'.

runtime/src/lib.rs
shubhamnarlawar: Yes. Makes sense. I have used `usize` instead of `u32` as match expects a `usize`-

`pub const BASE_TYPE_INDX: usize = 0;`
`pub const ARRAY_MEMBER_TYPE_INDX: usize = 1;`
`pub const ARRAY_SIZE_LSB_INDX: usize = 2;`
`pub const ARRAY_SIZE_MSB_INDX: usize = 3;`

I will push this snippet along with other changes.

runtime/src/lib.rs
```diff
@@ -19,84 +19,143 @@
...
+}
+
+pub fn calculate_type_priority(type_string: &str) -> i32 {
+    let mut priority = 0;
+    match type_string.chars().nth(1) {
+        Some('C') => priority = 4,
+        Some('B') => priority = 1,
+        Some('F') | Some('I') => priority = 8,
+        Some('A') => priority = 16,
+        _ => println!("Unsupported type"),
```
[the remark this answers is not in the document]
shubhamnarlawar: How about this?

_ => panic!("Unsupported type in mangled string"),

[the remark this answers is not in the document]
ruvi-d: I'd return a [std::result::Result](https://doc.rust-lang.org/std/result/enum.Result.html) from the function. The lib.rs side can check the return and panic if required.

shubhamnarlawar:
> @shubhamnarlawar other general review comments:
> 
>     1. Shall we move the logic out of lib.rs and into separate modules? Let's try to keep lib.rs as thin as possible
> 
>     2. Add some Rust unit tests for the new functions
> 
>     3. Do we need to handle unbounded size arrays? I believe that on the BIR side unbounded is handled with a `-1` size. How are we handling that?

For point 3, we have not planned handling unbounded size arrays in this sprint. We need to discuss it in Monday's meeting - 

Some quick thoughts to discuss - 
`What happens when one typecast such an array to a type that has a fixed length?  What happens if the fixed length array is larger than the current length of the variable length array?  What happens if it is smaller?`

runtime/src/lib.rs
```diff
@@ -19,84 +19,143 @@
...
+ * Index 2 and 3 - represents bits followed by its size
+ * */
 #[no_mangle]
-pub extern "C" fn is_same_type(src_type : *const c_char, dest_type : *const c_char) -> bool
-{
-  return src_type == dest_type;
+pub extern "C" fn is_same_type(src_type: *const c_char, dest_type: *const c_char) -> bool {
+    //Conversion to CStr
+    let source_cstr: &CStr = unsafe { CStr::from_ptr(src_type) };
+    let dest_cstr: &CStr = unsafe { CStr::from_ptr(dest_type) };
```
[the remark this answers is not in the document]
shubhamnarlawar: Done.

runtime/src/lib.rs
```diff
@@ -19,84 +19,143 @@
...
+    return size;
+}
+
+pub fn calculate_type_priority(type_string: &str) -> i32 {
+    let mut priority = 0;
+    match type_string.chars().nth(1) {
+        Some('C') => priority = 4,
+        Some('B') => priority = 1,
+        Some('F') | Some('I') => priority = 8,
+        Some('A') => priority = 16,
```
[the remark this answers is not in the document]
shubhamnarlawar: Done. Please check it!

runtime/src/lib.rs
```diff
@@ -19,84 +19,143 @@
...
+            if src_priority > dest_priority {
+                return false;
+            }
+            let src_size: i32 = compute_size(&source);
+            let dest_size: i32 = compute_size(&destination);
+            if src_size > dest_size {
+                return false;
+            }
+        }
+        _ => println!("Unsupported data structure"),
```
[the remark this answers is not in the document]
shubhamnarlawar: Sure. Done.

runtime/src/lib.rs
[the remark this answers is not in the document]
shubhamnarlawar: Done.

runtime/src/lib.rs
```diff
@@ -19,84 +19,175 @@
...
+ * Type Notations -
+ * C represents character
+ * B represents boolean
+ * F represents float
+ * I represents integer
+ * X represents any*/
+pub fn calculate_type_priority(type_string: &str) -> Result<i32, &'static str> {
+    let priority: i32;
+    match type_string.chars().nth(ARRAY_MEMBER_TYPE_INDX) {
+        Some('C') | Some('B') => {
```
manuranga, line 60: this should be descriptive.

shubhamnarlawar: I have added comment at top of this function- 

`/*`
 `* Computes size of the type`
 `* Type Notations -`
 `* C represents character`
 `* B represents boolean`
 `* F represents float`
 `* I represents integer`
 `* S represents string`
 `* X represents any*/`

Is using a macro suggested here ? Like - 

`macro_rules! Char { () => {'C'}; }`

and In the match case, it will be - `Some(Char!())`

WDYT?

shubhamnarlawar: @manuranga - In yesterday's call, you suggested to use enum. But here at some places, two entities require same value - 

for example - `Some('C') | Some('B')` - here both characer and boolean requires a value of `1`.

Can you share example to make it more descriptive? You mentioned that you have known in other languages in C++ to make this match case more descriptive. Can you please share here?

manuranga: Hi looks like rust doesn't support this out of the box yet https://github.com/rust-lang/rfcs/pull/3040
But it seems like there is a common way to do this : https://stackoverflow.com/a/28029279
Please try to use this approach

shubhamnarlawar: From above link - I can write enum as below - 

`enum TYPE_SIZE {`
`    CHAR = 1,`
`    INT = 8,`
`    STRING = 12,`
`    ANY = 16,`
`}`
`and so on`

but for boolean data type with size `1 byte`, it also need `1`
similarly, float needs `8`. Here we have repetitive values. How would enum works in this case?

manuranga: Hi, I am suggesting not to have unnamed constants like `C` and `X`. This is the same reason we use constants instead inlined numbers https://stackoverflow.com/a/47902.

I was not suggesting to have priorities be encoded into to enum.

shubhamnarlawar: Got it. I misinterpreted it.

How about using below constants instead of just `C` and `X`? - 

`pub const CHARACTER: char = 'C';`
`pub const BOOLEAN: char = 'B';`
`pub const INTEGER: char = 'I';`
`pub const FLOAT: char = 'F';`
`pub const STRING: char = 'S';`
`pub const ANY: char = 'X';`
`pub const ARRAY: char = 'A';`

runtime/src/lib.rs
```diff
@@ -19,84 +19,193 @@
...
+    if source.chars().nth(BASE_TYPE_INDX) != destination.chars().nth(BASE_TYPE_INDX) {
+        return false;
+    }
+    let mut src_type_size: Result<i32, &'static str> = Ok(0);
+    let mut dest_type_size: Result<i32, &'static str> = Ok(0);
+    match source.chars().nth(BASE_TYPE_INDX) {
+        Some('A') => {
+            if source.chars().nth(ARRAY_MEMBER_TYPE_INDX)
+                != destination.chars().nth(ARRAY_MEMBER_TYPE_INDX)
+            {
```
manuranga, line 127: is this line formatted using `rustfmt` ? brace seems to be on the wrong line.

runtime/src/lib.rs
```diff
@@ -19,84 +19,193 @@
...
+    //If type strings are same
+    if source == destination {
+        return true;
+    }
+    //Index 0 represents type of data structure
+    if source.chars().nth(BASE_TYPE_INDX) != destination.chars().nth(BASE_TYPE_INDX) {
+        return false;
+    }
+    let mut src_type_size: Result<i32, &'static str> = Ok(0);
+    let mut dest_type_size: Result<i32, &'static str> = Ok(0);
```
manuranga, line 122: In Rust, it's a best practice to use `mut` minimally. If you can avoid it, it's better.

runtime/src/lib.rs
```diff
@@ -19,84 +19,193 @@
...
+
+/*
+ * Returns size of the type
+ * Type Notations -
+ * C represents character
+ * B represents boolean
+ * F represents float
+ * I represents integer
+ * S represents string
+ * X represents any*/
```
manuranga, line 60: @shubhamnarlawar Please read https://ballerina.io/spec/lang/draft/v2020-12-17/#section_5.1.4
these are the type we need to support. We can start by selected subset, but there is no type called character.

runtime/src/lib.rs
manuranga: It took some time, but I was able to come up with following code. Please do the research on rust language best practices in the future.
```rust
extern crate num;
extern crate num_derive;

use num::FromPrimitive;
use num_derive::FromPrimitive;

const ARRAY_MEMBER_TYPE_INDEX: usize = 2;

#[derive(Debug, PartialEq, FromPrimitive)]
#[repr(u32)]
enum BalType {
    Nil = 'N' as u32,
    Boolean = 'B' as u32,
    Int = 'I' as u32,
    Float = 'F' as u32,
    Decimal = 'D' as u32,
    String = 'S' as u32,
    Array = 'A' as u32,
    Any = 'X' as u32,
}

pub fn type_size(type_string: &str) -> i32 {
    let type_char = type_string.chars().nth(ARRAY_MEMBER_TYPE_INDEX)
                                      .unwrap_or_else(|| panic!("illegle type discripor '{}', wrong length", type_string));
    let bal_type = FromPrimitive::from_u32(type_char as u32)
                                      .unwrap_or_else(|| panic!("illegle type tag '{}' in type discripor '{}'", type_char, type_string));
    match bal_type {
        BalType::Boolean => { 1 }
        BalType::Int | BalType::Float => { 8 }
        BalType::String => { 12 }
        BalType::Any => { 16 }
        _ => { unimplemented!("type_size for '{:?}'", bal_type) }
    }
}
```

Cargo.toml
```toml
[dependencies]
num = "0.3"
num-traits = "0.2"
num-derive = "0.3"
```

runtime/src/lib.rs
shubhamnarlawar: I am using `cargo-fmt`. I tried with `rustfmt` but getting same result. Let me reformat it once again!

runtime/src/lib.rs
shubhamnarlawar: Got it! Sure

runtime/src/lib.rs
shubhamnarlawar: Thank you for the snippet! Sure. I will follow it in next tasks!

Let me modify other functions to follow above points!

runtime/src/lib.rs
shubhamnarlawar: Got it. From the above enum you wrote, we can treat it as a subset!

ruvi-d:
@shubhamnarlawar remember to create a separate `type_checker.rs` file similar to how we created a new file for the map implementation.

shubhamnarlawar:
> @shubhamnarlawar remember to create a separate `type_checker.rs` file similar to how we created a new file for the map implementation.

I have moved all functions except `is_same_type()` and tests to `type_checker.rs`. Please check it!

shubhamnarlawar:
Is this PR ready to merge now?

ruvi-d:
> Is this PR ready to merge now?

You will have to fix the conflicts with master. When the PR is ready to review, you can mark my request for change complete and request for review again.

shubhamnarlawar:
> > Is this PR ready to merge now?
> 
> You will have to fix the conflicts with master. When the PR is ready to review, you can mark my request for change complete and request for review again.

I have resolved the conflict with master. No merge conflict now!
I have re-requested you for a review.

shubhamnarlawar:
@ruvi-d - Please give suggestions if any on latest changes pushed.

ruvi-d (changes requested):
Much better now with the Rust unit tests.
Just one more suggestion. Move the is_same_type function to the type_checker module. Change it's input args to take Rust strings. On the lib.rs, only do the String construction and directly pass it to type_checker module. Similar to how the Map implementation has been done. 
The idea here is to keep lib.rs as thin as possible, doing only the data type conversions. The actual logic will be kept on a separate, reusable(independent) Rust module.

shubhamnarlawar:
> Much better now with the Rust unit tests.
> Just one more suggestion. Move the is_same_type function to the type_checker module. Change it's input args to take Rust strings. On the lib.rs, only do the String construction and directly pass it to type_checker module. Similar to how the Map implementation has been done.
> The idea here is to keep lib.rs as thin as possible, doing only the data type conversions. The actual logic will be kept on a separate, reusable(independent) Rust module.

Done. Please check it.

I have moved logic of `is_same_type()` into a separate function called `same_type()` in `type_checker.rs` Now, the code is much cleaner.

ruvi-d (approved):
I am OK to merge. Best if @manuranga can double check the Rust lang specific stuff.

manuranga:
Still little bit unsure about the algo, but we'll catch if there is any issues with lit tests later.

Looks good to merge.

shubhamnarlawar:
Sounds good. Can you mark it as review complete and approve?

Kishanthan (approved):

manuranga:
@gimantha please send a pr to nBallerina updating comments with issue numbers when done.

PR #111, opened 2021-02-18
Add the nil type support

### LINKED ISSUE #94: Support for nil type
**Description:**
$subject.
This issue is a subtype of https://github.com/ballerina-platform/nballerina/issues/74

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

KavinduZoysa:
## Purpose
$subject

Fixes #94

For following basic program, [LLVM IR](https://github.com/KavinduZoysa/llvm/blob/assembly/main.ll) and [assembly file](https://github.com/KavinduZoysa/llvm/blob/assembly/assembly_file) are uploaded.

```
public function main() {
    () s = foo();
}

function foo() returns () {
    return ();
}
```

## Goals
Support for nil-type-descriptor

## Approach
Create a global variable for nul value and change the places where constLoad, return, functionGen.
`@bal_nil_value = global i8* null, align 8`

## Test environment
ubuntu 16.04
llvm-11.0.1


compiler/llvmFunctionGen.cpp
```diff
@@ -93,6 +93,11 @@ LLVMTypeRef BIRFunction::getLLVMFuncRetTypeRefOfType(VarDecl *vDecl) {
   case TYPE_TAG_CHAR_STRING:
   case TYPE_TAG_STRING:
     return LLVMPointerType(LLVMInt8Type(), 0);
+  case TYPE_TAG_NIL:
+    if (funcName == "main") {
+      return LLVMVoidType();
```
ruvi-d, line 98: The special casing of `_bal_result` at line num:71-83, should move inside this block right? otherwise if there is a `_bal_result` defined all functions that return void will return int instead.

compiler/llvmConstantLoadInsn.cpp
```diff
@@ -75,7 +75,14 @@ void ConstantLoadInsn::translate(LLVMModuleRef &modRef) {
       break;
     }
     case TYPE_TAG_NIL: {
-      return;
+      // check for the main function and () is assigned to 0%
+      if (getFunction() -> getName() == "main" && lhsOp -> getVarDecl() -> getVarName() == "%0") {
```
ruvi-d, line 79: make `main` a constexpr char just like `BAL_NIL_VALUE`.
also formatting (spacing) is off here .

compiler/llvmFunctionGen.cpp
```diff
@@ -93,6 +93,11 @@ LLVMTypeRef BIRFunction::getLLVMFuncRetTypeRefOfType(VarDecl *vDecl) {
...
+    }
+    return LLVMPointerType(LLVMInt8Type(), 0);
```
ruvi-d, line 100: let's use `getLLVMTypeRefOfType()` function to get the LLVMTypeRef instead of replicating the switch case here again

compiler/llvmConstantLoadInsn.cpp
```diff
@@ -75,7 +75,14 @@ void ConstantLoadInsn::translate(LLVMModuleRef &modRef) {
...
+        return; 
+      }
+      LLVMValueRef constTempRef =  getPkgAddress()->getGlobalVarRefUsingId(BAL_NIL_VALUE);
+      string tempName = BAL_NIL_VALUE + "_temp";
```
ruvi-d, line 83: What if there are more than one nil type variables in the function? this would then create two constLoad instructions with the same tempName. lhsOp name + _temp would be better right?

compiler/llvmPackageGen.cpp
```diff
@@ -45,9 +45,18 @@ void BIRPackage::translate(LLVMModuleRef &modRef) {
...
   }
 
+  // create global var for nil value  
+  Constant *nullValue = Constant::getNullValue(unwrap(LLVMPointerType(LLVMInt8Type(), 0)));
+  GlobalVariable *gVar = new GlobalVariable(
+        *unwrap(modRef), unwrap(LLVMPointerType(LLVMInt8Type(), 0)), false,
+        GlobalValue::ExternalLinkage, nullValue, BAL_NIL_VALUE.c_str(), 0);
+  gVar->setAlignment(Align(8));
+  LLVMValueRef globVarRef = wrap(gVar);
+  globalVarRefs.insert({BAL_NIL_VALUE, globVarRef});
```
ruvi-d, line 58: This might be a problem some time later when we are going link multiple translation units because of the `GlobalValue::ExternalLinkage` used here. Better change it to internal linkage.

compiler/llvmPackageGen.cpp
```diff
@@ -75,7 +84,8 @@ void BIRPackage::translate(LLVMModuleRef &modRef) {
       isVarArg = true;
 
     if (birFunc->getReturnVar())
-      retType = birFunc->getLLVMFuncRetTypeRefOfType(birFunc->getReturnVar());
+      retType = birFunc->getLLVMFuncRetTypeRefOfType(birFunc->getReturnVar(), 
+                                                                                                              birFunc->getName());
```
ruvi-d, line 88: Update formatting

compiler/llvmReturnInsn.cpp
```diff
@@ -24,29 +24,28 @@ void ReturnInsn::translate(LLVMModuleRef &modRef) {
       }
     }
   }
-  if (funcObj->getName() == "main" && returnVarDecl) {
```
ruvi-d, line 27: Prefer early exits pattern to prevent multiple nested if/else blocks. Ref: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

test/types/nil/nil.bal
```diff
@@ -0,0 +1,15 @@
+// RUN: JAVA_HOME=%java_path %testRunScript %s %nballerinacc | filecheck %s
+
+public function main() {
+  () nilVal = bar();
+}
+
+function foo() {
+  () nilVal = ();
+}
```
ruvi-d, line 9: Not using `foo()`?

ruvi-d (changes requested):
Please see rview comments

compiler/llvmFunctionGen.cpp
KavinduZoysa: If `_bal_result` is there, the logic at L71-L83 changes the target type to `int`. Therefore it returns the correct type.

compiler/BIR.h
```diff
@@ -28,6 +28,9 @@
 using namespace std;
 using namespace llvm;
 
+// define global nil value name
+const string BAL_NIL_VALUE = "bal_nil_value";
+
```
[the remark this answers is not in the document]
KavinduZoysa: After changing this I got an error when the string is concatenated as `BAL_NIL_VALUE + "_temp";`. Then I used `strcpy`, but still getting an error because it is a constexpr. Any suggestions to fix this?

compiler/llvmConstantLoadInsn.cpp
KavinduZoysa: Fixed.

compiler/llvmFunctionGen.cpp
KavinduZoysa: We need to make the `main` function a special case right? In `getLLVMTypeRefOfType()` we are not doing such a thing and default cases are different.

compiler/llvmConstantLoadInsn.cpp
KavinduZoysa: Yes, fixed

compiler/llvmPackageGen.cpp
KavinduZoysa: Fixed

compiler/llvmReturnInsn.cpp
KavinduZoysa: Fixed

test/types/nil/nil.bal
KavinduZoysa: I did not get this comment.

compiler/llvmPackageGen.cpp
KavinduZoysa: Fixed

compiler/BIR.h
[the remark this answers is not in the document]
ruvi-d: I'd create a new string from BAL_NIL_VALUE to concat, like `std::string bal_nil(BAL_NIL_VALUE); //bal_nil + "_temp"`

compiler/llvmPackageGen.cpp
```diff
@@ -45,9 +45,18 @@ void BIRPackage::translate(LLVMModuleRef &modRef) {
...
-      globalVarRefs.insert({globVar->getVarName(), globVarRef});
+      globalVarRefs.insert({varName, globVarRef});
   }
 
+  // create global var for nil value  
+  Constant *nullValue = Constant::getNullValue(unwrap(LLVMPointerType(LLVMInt8Type(), 0)));
+  GlobalVariable *gVar = new GlobalVariable(
+        *unwrap(modRef), unwrap(LLVMPointerType(LLVMInt8Type(), 0)), false,
+        GlobalValue::InternalLinkage, nullValue, BAL_NIL_VALUE.c_str(), 0);
+  gVar->setAlignment(Align(8));
```
manuranga, line 56: Please research and see if the alignment is needed, or can llvm figure it out automatically .

compiler/llvmFunctionGen.cpp
ruvi-d: I mean for an example like this:
```
int _bal_result = 0;
public function main() {
  () nilVal = bar();
}

function foo() {
  () nilVal = ();
}
```
foo()'s return would incorrectly change to int type right?

compiler/llvmFunctionGen.cpp
ruvi-d: I'd rather we use getLLVMTypeRefOfType() and only override with `(funcName == "main")` for `TYPE_TAG_NIL`. Otherwise we would be duplicating the logic and might miss to update one in the future.

compiler/llvmPackageGen.cpp
KavinduZoysa: I think we do not need that, please refer https://llvm.org/docs/LangRef.html#id1694

compiler/BIR.h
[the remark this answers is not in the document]
KavinduZoysa: Thank you, fixed it.

compiler/llvmFunctionGen.cpp
KavinduZoysa: yes, fixed.

compiler/llvmFunctionGen.cpp
KavinduZoysa: I am still confused about how to use`getLLVMTypeRefOfType`. 
Will it work in below case?
```public function main() {
  () nilVal = bar();
}```

Also in both methods, default cases are different.

Comment (ruvi-d) [reply compiler/llvmFunctionGen.cpp]: how about some thing like this
```
  if(typeTag != TYPE_TAG_NIL)
    return getLLVMTypeRefOfType(vDecl->getTypeDecl());
  if (funcName == "main") {
    return LLVMVoidType();
  }
  return LLVMPointerType(LLVMInt8Type(), 0);
```
that is without the `_bal_result` logic

File: test/types/nil/nil.bal (hunk 8 above)
Comment (ruvi-d) [reply test/types/nil/nil.bal:9]: sorry i missed your reply here.
I meant that you are defining the foo() function here, but because it is not being used. So the codegen might be incorrect but the test will still pass. Shall we at least call it from the main function?

Comment (KavinduZoysa) [reply test/types/nil/nil.bal:10]: Yes, changed it.

File: compiler/llvmFunctionGen.cpp (hunk 3 above)
Comment (KavinduZoysa) [reply compiler/llvmFunctionGen.cpp]: I have changed it in a different way, could you please check.

**Review by ruvi-d — APPROVED** [review]
Good to merge

File: compiler/llvmFunctionGen.cpp [hunk 11]
```diff
@@ -93,6 +80,22 @@ LLVMTypeRef BIRFunction::getLLVMFuncRetTypeRefOfType(VarDecl *vDecl) {
   case TYPE_TAG_CHAR_STRING:
   case TYPE_TAG_STRING:
     return LLVMPointerType(LLVMInt8Type(), 0);
+  case TYPE_TAG_NIL : {
+    if (funcName != MAIN_FUNCTION_NAME) {
+      return LLVMPointerType(LLVMInt8Type(), 0);
+    }
```
ayonam, line 86: Why would you want to return an i8* as the return type for any function that is not main but returns nil?  Why not the same as what we do for main, i.e., the global NIL value?  Do we expect to do anything with this nil value?  If not, then it shouldn't matter if multiple functions point to the same value as their return value, right?

@manuranga what do you think?

ayonam (changes requested):

compiler/BIR.h
```diff
@@ -804,7 +808,7 @@ class BIRFunction : public BIRNode {
   LLVMValueRef getLocalToTempVar(Operand *op);
   void translateFunctionBody(LLVMModuleRef &modRef);
   void patchInsn(Function *llvnFun);
-  LLVMTypeRef getLLVMFuncRetTypeRefOfType(VarDecl *vDecl);
+  LLVMTypeRef getLLVMFuncRetTypeRefOfType(VarDecl *vDecl, string funcName);
```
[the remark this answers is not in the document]
KavinduZoysa: Yes Ayonam, we don't need that, fixed it.

compiler/llvmConstantLoadInsn.cpp
```diff
@@ -75,7 +75,15 @@ void ConstantLoadInsn::translate(LLVMModuleRef &modRef) {
...
     case TYPE_TAG_NIL: {
-      return;
+      string lhsOpName = lhsOp->getVarDecl()->getVarName();
+      // check for the main function and () is assigned to 0%
+      if (getFunction()->getName() == MAIN_FUNCTION_NAME && lhsOpName == "%0") {
+        return; 
+      }
+      LLVMValueRef constTempRef =  getPkgAddress()->getGlobalVarRefUsingId(BAL_NIL_VALUE);
+      string tempName = std::string(BAL_NIL_VALUE) + "_" + lhsOpName + "_temp";
+      constRef = LLVMBuildLoad(builder, constTempRef, tempName.c_str());
```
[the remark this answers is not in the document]
KavinduZoysa: `constRef` is stored here, https://github.com/ballerina-platform/nballerina/blob/976d306f0dfac18f7a0497b7c88efea37eea831f/compiler/llvmConstantLoadInsn.cpp#L85

This is the thing that was meant in this comment?

compiler/llvmReturnInsn.cpp
```diff
@@ -24,29 +24,29 @@ void ReturnInsn::translate(LLVMModuleRef &modRef) {
...
+  if (!returnVarDecl) {
+    LLVMBuildRetVoid(builder);
+    return;
+  }
+
+  if (builder && funcObj) {
+    LLVMValueRef lhsRef =
+        funcObj->getLocalVarRefUsingId(returnVarDecl->getVarName());
+    if (lhsRef) {
+      return;
```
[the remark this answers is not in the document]
KavinduZoysa: That is correct Ayonam, this check is redundant, therefore removed it.

compiler/llvmFunctionGen.cpp
KavinduZoysa: [Ballerina spec](https://htmlpreview.github.io/?https://raw.githubusercontent.com/ballerina-platform/ballerina-spec/master/lang/spec.html) says `An empty return-type-descriptor is equivalent to returns ()`. Also we have discussed the same in GitHub issue https://github.com/ballerina-platform/nballerina/issues/94#issuecomment-781046163.

Therefore all functions(except main) which do not return, implemented as returning ().

test/types/nil/nil.bal
```diff
@@ -0,0 +1,16 @@
+// RUN: JAVA_HOME=%java_path %testRunScript %s %nballerinacc | filecheck %s
+
+public function main() {
+  () nilVal = bar();
+  foo();
+}
+
+function foo() {
+  () nilVal = ();
+}
```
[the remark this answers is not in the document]
KavinduZoysa: Added a new function

compiler/llvmReturnInsn.cpp
```diff
@@ -24,29 +24,29 @@ void ReturnInsn::translate(LLVMModuleRef &modRef) {
...
-        lhsRef = getPkgAddress()->getGlobalVarRefUsingId(
-            returnVarDecl->getVarName());
 
-        LLVMValueRef retValRef = LLVMBuildLoad(builder, lhsRef, "retrun_temp");
-        if (retValRef)
-          LLVMBuildRet(builder, retValRef);
-      }
+  if (funcObj->getName() != MAIN_FUNCTION_NAME) {
+    LLVMValueRef retValueRef = LLVMBuildLoad(
+    builder, funcObj->getLocalVarRefUsingId("%0"), "retrun_temp");
```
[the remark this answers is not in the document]
KavinduZoysa: We do not have proper documentation for that, but it is kind of a convention in the language. The return value is represented  using `%0`, in the .ll file it becomes `%"%0"`  @manuranga please correct me if I am wrong.

compiler/llvmConstantLoadInsn.cpp
```diff
@@ -75,7 +75,15 @@ void ConstantLoadInsn::translate(LLVMModuleRef &modRef) {
       break;
     }
     case TYPE_TAG_NIL: {
-      return;
+      string lhsOpName = lhsOp->getVarDecl()->getVarName();
+      // check for the main function and () is assigned to 0%
+      if (getFunction()->getName() == MAIN_FUNCTION_NAME && lhsOpName == "%0") {
+        return; 
+      }
```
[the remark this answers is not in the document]
ruvi-d: Kavindu, sounds like we will have to use a function like BIRFunction->getReturnVar() to get the return variable and compare it with your lhs. Instead of assuming that the return variable is %0.

compiler/llvmReturnInsn.cpp
```diff
@@ -24,29 +24,26 @@ void ReturnInsn::translate(LLVMModuleRef &modRef) {
...
+
+  if (!returnVarDecl) {
+    LLVMBuildRetVoid(builder);
+    return;
+  }
+
+  if (builder && funcObj) {
+    LLVMValueRef lhsRef =
+        funcObj->getLocalVarRefUsingId(returnVarDecl->getVarName());
+    lhsRef = getPkgAddress()->getGlobalVarRefUsingId(returnVarDecl->getVarName());
```
ruvi-d, line 43: We are setting the `lhsRef` from the `getLocalVarRefUsingId` function and then immediately reassigning it with the return from `getGlobalVarRefUsingId`? Are we missing a `if(lhsRef)` before line 43?

compiler/llvmConstantLoadInsn.cpp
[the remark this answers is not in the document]
KavinduZoysa: Changed

compiler/llvmReturnInsn.cpp
KavinduZoysa: Yes, fixed.

compiler/llvmReturnInsn.cpp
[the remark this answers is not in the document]
KavinduZoysa: This is changed to use `getReturnVar()` instead of `%0`

ruvi-d (approved):
Good to merge

compiler/llvmConstantLoadInsn.cpp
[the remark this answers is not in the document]
ayonam: How do I see the changed code here?  I don't see any call to getReturnVar().

compiler/llvmConstantLoadInsn.cpp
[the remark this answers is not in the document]
ayonam: Yes, this is fine.  Sorry, I missed that.

compiler/llvmReturnInsn.cpp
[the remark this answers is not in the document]
ayonam: Again I can't see the changes made.  Also, have we changed that "retrun_temp" to "return_temp"?  Or may be "return_val_temp" or "retval_temp" is a more apt name.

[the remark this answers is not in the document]
KavinduZoysa: @ayonam, please go to the list of `Files Changed` navigate to the file. The new change will not appear here.

compiler/llvmConstantLoadInsn.cpp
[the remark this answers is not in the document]
KavinduZoysa: @ayonam, please go to the list of `Files Changed` navigate to the file. The new change will not appear here.

compiler/llvmFunctionGen.cpp
ayonam: But for main() we are returning the global NIL value, right?  So we should do the same for these functions as well.  Why are we returning an i8*?

compiler/llvmConstantLoadInsn.cpp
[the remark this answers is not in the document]
ayonam: Okay.  Got it.  Thanks.

compiler/llvmFunctionGen.cpp
KavinduZoysa: In the main function, we are returning void, because we have to wrap ballerina main with native main https://github.com/ballerina-platform/nballerina/issues/94#issuecomment-781080663. Here we are returning the type of global nil value which is i8*. @ayonam, as I understood you are asking the reason to return nil for all functions except main right?

ayonam: Okay, I got it, but still this code does not handle the case where a user wants to return a value from main() through the _bal_result global variable.  I think the logic of the deleted lines 72 to 81 needs to be implemented here.

ayonam: Sorry, didn't notice. That code is there right below this comment.

ayonam (approved):

Kishanthan (commented):
The sources (both Ballerina and cpp) aren't formatted according to the convention

Kishanthan (approved):

PR #164, opened 2021-03-10
Enable LIT tests in windows

### LINKED ISSUE #101: Enable LIT tests for windows
**Description:**
Currently, the generated .ll files in windows are not correct, therefore LIT tests are not enabled. Those need to be enabled after fixing #95

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**
#95

KavinduZoysa:
## Purpose
$subject

Fixes #101


CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ target_link_libraries(nballerinacc PRIVATE ${llvm_libs})
 # Set extra warnings for nballerinacc
 if (MSVC)
     set(DEBUG_OPTIONS /W4)
-    set(RELEASE_OPTIONS /W3 /WX)
+    set(RELEASE_OPTIONS /W4)
```
ruvi-d, line 44: Any reason to disable treat warnings as errors (/WX) flag?

compiler/codegen/TypeCastInsn.cpp
```diff
@@ -24,6 +24,9 @@
 #include "Variable.h"
 #include "llvm-c/Core.h"
 #include "llvm/IR/Constants.h"
+#ifndef unix
+#define __attribute__(unused)
+#endif
```
ruvi-d, line 29: You can use `[[maybe_unused]]` instead of `__attribute__(unused)` like in line 36. maybe_unused is a C++ lang spec.

ruvi-d (commented):
Added a few minor comments

CMakeLists.txt
KavinduZoysa: In windows, we build nballerina as a release(I think the reason we have to do this is, llvm was build as a release), therefore warnings are treated as errors. Fixing those is out of this scope, so I disabled it, lets enable after fixing those warnings.

compiler/codegen/TypeCastInsn.cpp
KavinduZoysa: Sure, will update

KavinduZoysa: Sure, will update

KavinduZoysa: Changed

ruvi-d (approved):

uthaiyashankar (approved):

Issue #158, opened 2021-03-08
Refactor BIRReader.h/.cpp into smaller files


ruvi-d:
**Description:**

Refactor BIRReader.h and cpp files into multiple smaller files similar to what #114 did for BIR.h.
These new files can be moved in to a sub directory inside the compiler folder.

**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Kishanthan:
@heshanpadmasiri

PR #180, opened 2021-03-15
Fix memory leaks

### LINKED ISSUE #115: Fix memory leaks
**Description:**
Doing a static analysis or enabling sanitizers show that that is a large number of memory leaks in the LLVM codegen classes. Suggest that we fix these leaks using smart pointers.

**Suggested Labels:**
Priority/High, Type/Bug, Points/2

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

ruvi-d:
## Purpose
Refactor all classes to use pointers more judiciously. 

Resolves #115 
Not in scope : #158  

## Approach
* Avoid using pointers for class member
* Return object references instead of pointers when it can't be nullptr
* Use of const keyword to indicate and enforce immutability 
* Use of unique or shared pointers instead of raw pointers


ruvi-d:
@ayonam kind reminder to review this PR

compiler/include/BasicBlock.h
```diff
@@ -32,30 +32,33 @@ class Function;
 class TerminatorInsn;
 class NonTerminatorInsn;
 
-class BasicBlock : public PackageNode, public Debuggable, public Translatable {
+class BasicBlock : public Debuggable, public Translatable {
```
ayonam, line 35: Why do we no longer derive from PackageNode?

compiler/include/ConstantLoad.h
```diff
@@ -29,35 +30,21 @@ namespace nballerina {
...
-        int intValue;
-        double floatValue;
-        bool boolValue;
-        std::string *strValue;
-        value() {}
-        value(int x) : intValue(x) {}
-        value(float x) : floatValue(x) {}
-        value(bool x) : boolValue(x) {}
-        value(std::string *x) : strValue(x) {}
-    } val;
```
ayonam, line 42: Since only one value can be active at a time, why not have this as a union?  The different constructors for different value types can still be there and they could invoke the relevant constructor for value().

compiler/include/Function.h
```diff
@@ -32,69 +35,67 @@ namespace nballerina {
...
 class FunctionParam;
-class Variable;
-class InvokableType;
-class RestParam;
+class InvocableType;
 class Type;
 class Package;
 
-class Function : public PackageNode, public Debuggable, public Translatable {
+class Function : public Debuggable, public Translatable {
```
ayonam, line 42: Again here we don't derive from PackageNode. Shouldn't we be having the Package information with the Function?  We may need to access it for various reasons.  Just my thoughts.

compiler/include/FunctionCallInsn.h
```diff
@@ -30,18 +30,14 @@ class FunctionCallInsn : public TerminatorInsn {
...
     std::string functionName;
     int argCount;
-    std::vector<Operand *> argsList;
+    std::vector<Operand> argsList;
 
   public:
     FunctionCallInsn() = delete;
-    FunctionCallInsn(bool funcVirtual, std::string _funcName, int argNumber, BasicBlock *nextBB, Operand *lhsOp,
-                     std::vector<Operand *> &fnArgs, BasicBlock *currentBB);
+    FunctionCallInsn(std::string _funcName, int argNumber, std::shared_ptr<BasicBlock> nextBB, const Operand &lhs,
```
ayonam, line 37: What was the relevance of funcVirtual earlier?  Why have we dropped it?  These were added to these classes basis the fields in the BIR.  So unless they are no longer relevant in the BIR, we should not be dropping them.

compiler/include/FunctionParam.h
```diff
@@ -20,21 +20,22 @@
...
 
 class FunctionParam : public Operand {
   private:
-    Type *type;
+    Type type;
 
   public:
     FunctionParam() = delete;
-    FunctionParam(std::string name) : Operand(std::move(name), ARG_VAR_KIND), type(nullptr) {}
+    FunctionParam(Operand paramOp, Type type) : Operand(std::move(paramOp)), type(std::move(type)) {}
```
ayonam, line 35: I'm assuming that the name of the operand has been captured while creating the paramOp.

compiler/include/Location.h
```diff
@@ -26,21 +26,21 @@ namespace nballerina {
 class Location {
   private:
     std::string fileName;
-    const int sLine;
-    const int sCol;
-    const int eLine;
-    const int eCol;
+    int sLine;
```
ayonam, line 29: Shouldn't these be const here?  We don't expect them to change, do we?  Besides there are no set methods either.

compiler/include/Package.h
```diff
@@ -19,64 +19,73 @@
...
 
 class Package : public Translatable {
   private:
+    inline static const std::string BAL_NIL_VALUE = "bal_nil_value";
+    inline static const std::string STRING_TABLE_NAME = "__string_table_ptr";
     std::string org;
     std::string name;
     std::string version;
     std::string sourceFileName;
-    std::vector<Function *> functions;
```
ayonam, line 42: I'm assuming that the pointers to the Function objects are being stored in the functionLookUp map and hence this vector is redundant.

compiler/include/Package.h
```diff
@@ -19,64 +19,73 @@
...
-    llvm::StructType *boxType;
-    llvm::StringTableBuilder *strBuilder;
+    std::map<std::string, std::shared_ptr<Function>> functionLookUp;
+    std::unique_ptr<llvm::StructType> boxType;
+    std::unique_ptr<llvm::StringTableBuilder> strBuilder;
     std::map<std::string, std::vector<LLVMValueRef>> structElementStoreInst;
     std::map<std::string, LLVMValueRef> functionRefs;
+    std::vector<std::unique_ptr<llvm::GlobalVariable>> globalStringValues;
     void applyStringOffsetRelocations(LLVMModuleRef &modRef);
-    inline static const std::string BAL_NIL_VALUE = "bal_nil_value";
```
ayonam, line 51: Are we now storing the NIL value elsewhere?

compiler/include/RestParam.h
```diff
@@ -24,14 +24,9 @@
 namespace nballerina {
 
 class RestParam {
-  private:
-    std::string name;
-
```
ayonam, line 29: Don't the RestParams have a name anymore?  I thought the BIR sent a name here.

compiler/include/TypeCastInsn.h
```diff
@@ -30,15 +30,13 @@ class Operand;
 
 class TypeCastInsn : public NonTerminatorInsn {
   private:
-    Operand *rhsOp;
-    LLVMValueRef getIsSameTypeDeclaration(LLVMModuleRef &modRef, Package *pkg, LLVMValueRef lhsRef,
-                                          LLVMValueRef rhsRef);
```
ayonam, line 35: In this function, are we not using the Package anymore?

compiler/include/Types.h
```diff
@@ -83,29 +85,19 @@ class Type {
   private:
     TypeTag type;
     std::string name;
-    const int flags;
+    int flags;
```
ayonam, line 88: Do we expect to change the flags here?  These are directly obtained from the BIR.  So I don't think we will be changing them once they are set.  Shouldn't this should remain a const?

compiler/include/interfaces/AbstractInstruction.h
```diff
@@ -70,20 +71,22 @@ enum InstructionKind {
     INSTRUCTION_KIND_BINARY_BITWISE_UNSIGNED_RIGHT_SHIFT
 };
 
-class AbstractInstruction : public Debuggable, public PackageNode {
+class AbstractInstruction : public Debuggable {
```
ayonam, line 74: Again, why is the Package not made accessible to these classes?  There are times, when we may need it for the LLVM code generation.  I thought Package was the BIR synonym the LLVM Module.

ayonam (changes requested):

compiler/codegen/InvokableType.cpp
```diff
@@ -1,34 +0,0 @@
...
-
-InvokableType::InvokableType(std::vector<Type *> paramTy, Type *retTy)
-    : paramTypes(std::move(paramTy)), restType(nullptr), returnType(retTy) {}
-
-const Type *InvokableType::getReturnType() { return returnType; }
-const Type *InvokableType::getRestType() { return restType; }
-Type *InvokableType::getParamType(int i) { return paramTypes[i]; }
-size_t InvokableType::getParamTypeCount() { return paramTypes.size(); }
-
-} // namespace nballerina
```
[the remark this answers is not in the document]
ruvi-d: Renamed the class from `InvokableType` to `InvocableType`

compiler/codegen/Types.cpp
```diff
@@ -44,9 +54,4 @@ const char *Type::getNameOfType(TypeTag typeTag) {
     }
 }
 
-MapTypeDecl::MapTypeDecl(TypeTag type, std::string name, int flags, TypeTag memberType)
-    : Type{type, std::move(name), flags}, memberType{memberType} {}
-
-TypeTag MapTypeDecl::getMemberTypeTag() { return memberType; }
-
```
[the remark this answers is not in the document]
ruvi-d: Moved the one extra class member supported by MapTypeDecl (memberType) to the Type base class as an optional member. The overhead of introducing MapTypeDecl just to support one additional trivial member seemed unnecessary. If we had used MapTypeDecl, then we would have had to use heap allocated pointers to deal with Type objects and then incur extra overhead for the de-referencing.

compiler/include/BasicBlock.h
ruvi-d: The PackageNode interface only had a getPackage() method and that method implementation was moved to the respective classes. It was only used as a private method by their respective classes. Exposing this method as a interface's public method didn't net any benefit.

compiler/include/ConstantLoad.h
ruvi-d: Yes, actually a std::variant (type safe version of union) would make sense here. Will update.

compiler/include/Function.h
ruvi-d: Commented here: https://github.com/ballerina-platform/nballerina/pull/180#discussion_r597385022

compiler/include/FunctionCallInsn.h
ruvi-d: Currently when we are parsing BIR fields, we are leaving some unused and passing others to the CodeGen class constructors. But some of these fields were being unused in the CodeGen classes as well. So, to make things consistent and easily understand for future developers, I've removed all the unused params from class constructors and marked them as `unused` in the BIRReader. Then by looking at the BIRReader code you know exactly which fields are being used or ignored currently.

compiler/include/FunctionParam.h
ruvi-d: That's right

compiler/include/Package.h
ruvi-d: Yes

compiler/include/Location.h
ruvi-d: Using const member variables is discouraged since it makes it non trivial to copy and thus has adverse side effects like preventing moves. The recommendation I've seen is to keep member variables non-const and remove the setter methods instead.
https://www.reddit.com/r/cpp/comments/8wbeom/coding_guideline_avoid_const_member_variables/

compiler/include/Package.h
ruvi-d: It's just been moved a few lines above. The diff isn't showing that.

compiler/include/RestParam.h
ruvi-d: This change was also done for the same reason as mentioned previously about removing unused fields. 
https://github.com/ballerina-platform/nballerina/pull/180#discussion_r597387522

compiler/include/TypeCastInsn.h
ruvi-d: A reference to the parent package is already available to this class object. So unless we make this method Static, there was no need to pass a redundant package pointer

compiler/include/Types.h
ruvi-d: I am following the previously mentioned guidelines about const member variables
https://github.com/ballerina-platform/nballerina/pull/180#discussion_r597389997

compiler/include/interfaces/AbstractInstruction.h
ruvi-d: Answered in https://github.com/ballerina-platform/nballerina/pull/180#discussion_r597385022

compiler/include/ConstantLoad.h
ruvi-d: Push an update addressing this

ruvi-d:
@ayonam I've replied to your comments. Please check.

ruvi-d:
Rebased with master

compiler/codegen/Types.cpp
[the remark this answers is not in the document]
ayonam: Okay.  Got it.  That's fine.

compiler/include/BasicBlock.h
ayonam: Accepted.

compiler/include/FunctionCallInsn.h
ayonam: That sounds reasonable.

compiler/include/RestParam.h
ayonam: Got it.  Thanks,

ayonam (approved):

KavinduZoysa (approved):

PR #191, opened 2021-03-19
Box the data in any type

### LINKED ISSUE #190: Issue in returning any type
**Description:**
For any type, the smart pointer was created inside the stack, data is destroyed when returning. Therefore returned value does not contain the expected value.

**Steps to reproduce:**
function foo() returns any {
   return 32;
}

public function main() {
    int i = <int>foo(); // i != 32
}

KavinduZoysa:
## Purpose
$subject

Fixes #190

## Approach
Bring the data of any value to heap using box.

### DIFF
--- a/compiler/codegen/TypeCastInsn.cpp
+++ b/compiler/codegen/TypeCastInsn.cpp
@@ -122,9 +122,8 @@ void TypeCastInsn::translate([[maybe_unused]] LLVMModuleRef &modRef) {
             LLVMValueRef rhsTempOpRef = funcObj->getTempLocalVariable(rhsOp);
             LLVMValueRef *paramRefs = new LLVMValueRef[1];
             paramRefs[0] = rhsTempOpRef;
-            // LLVMValueRef boxIntFunc = boxInteger(modRef, rhsTempOpRef);
-            LLVMValueRef boxIntFunc = boxValue(modRef, rhsTempOpRef, rhsTypeTag);
-            rhsOpRef = LLVMBuildCall(builder, boxIntFunc, paramRefs, 1, "call");
+            LLVMValueRef boxValFunc = generateBoxValueFunc(modRef, LLVMTypeOf(rhsTempOpRef), rhsTypeTag);
+            rhsOpRef = LLVMBuildCall(builder, boxValFunc, paramRefs, 1, "call");
         }
         LLVMValueRef bitCastRes = LLVMBuildBitCast(builder, rhsOpRef, LLVMPointerType(LLVMInt8Type(), 0), "");
         LLVMBuildStore(builder, bitCastRes, elePtr2);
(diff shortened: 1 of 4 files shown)


test/casting/anyTypeReturn.bal
```diff
@@ -0,0 +1,32 @@
...
+function fooInt() returns any {
+    return 221;
+}
+
+function fooBool() returns any {
+    return true;
+}
+
+function fooFloat() returns any {
+    return 0.0;
```
ruvi-d, line 18: Change to something like 2.5, so that we can verify the return val

compiler/include/TypeCastInsn.h
```diff
@@ -39,6 +39,11 @@ class TypeCastInsn : public NonTerminatorInsn {
     TypeCastInsn(Operand *lOp, BasicBlock *currentBB, Operand *rOp, Type *tDecl, bool checkTypes);
     ~TypeCastInsn() = default;
     LLVMValueRef isSameType(LLVMModuleRef &modRef, LLVMValueRef lhsRef, LLVMValueRef rhsRef);
+    LLVMValueRef boxInteger(LLVMModuleRef &modRef, LLVMValueRef lhsRef);
+    LLVMValueRef boxBoolean(LLVMModuleRef &modRef, LLVMValueRef paramRef);
+    LLVMValueRef boxFloat(LLVMModuleRef &modRef, LLVMValueRef paramRef);
+    LLVMValueRef boxValue(LLVMModuleRef &modRef, LLVMValueRef paramRef, TypeTag typeTag);
```
ruvi-d, line 45: Make the new functions private

runtime/src/lib.rs
```diff
@@ -187,3 +187,18 @@ pub extern "C" fn map_store_int(ptr: *mut BalMapInt, key: *const c_char, member_
...
+
+#[no_mangle]
+pub extern "C" fn box_bal_float(val: f64) -> *mut f64 {
+    Box::into_raw(Box::new(val))
+}
+
+#[no_mangle]
+pub extern "C" fn box_bal_bool(val: i8) -> *mut i8 {
+    Box::into_raw(Box::new(val))
+}
```
ruvi-d, line 204: Need to add functions to free (deinit) the pointers

compiler/codegen/TypeCastInsn.cpp
```diff
@@ -117,13 +117,91 @@ void TypeCastInsn::translate([[maybe_unused]] LLVMModuleRef &modRef) {
...
+    }
+    case TYPE_TAG_BOOLEAN: {
+        return boxBoolean(modRef, paramRef);
+    }
+    default:
+        new LLVMValueRef();
+    }
+}
+
+bool TypeCastInsn::isBoxValueSupport(TypeTag typeTag) {
```
ruvi-d, line 194: make static

compiler/codegen/TypeCastInsn.cpp
```diff
@@ -117,13 +117,91 @@ void TypeCastInsn::translate([[maybe_unused]] LLVMModuleRef &modRef) {
...
+        return boxInteger(modRef, paramRef);
+    }
+    case TYPE_TAG_FLOAT: {
+        return boxFloat(modRef, paramRef);
+    }
+    case TYPE_TAG_BOOLEAN: {
+        return boxBoolean(modRef, paramRef);
+    }
+    default:
+        new LLVMValueRef();
```
ruvi-d, line 190: default should be an error

compiler/codegen/TypeCastInsn.cpp
```diff
@@ -117,13 +117,91 @@ void TypeCastInsn::translate([[maybe_unused]] LLVMModuleRef &modRef) {
...
+    if (boxBoolFuncRef != nullptr) {
+        return boxBoolFuncRef;
+    }
+    LLVMTypeRef *paramTypes = new LLVMTypeRef[1];
+    paramTypes[0] = LLVMTypeOf(paramRef);
+    LLVMTypeRef funcType = LLVMFunctionType(LLVMPointerType(LLVMInt8Type(), 0), paramTypes, 1, 0);
+    boxBoolFuncRef = LLVMAddFunction(modRef, boxBoolChar, funcType);
+    getPackage()->addFunctionRef(boxBoolChar, boxBoolFuncRef);
+    return boxBoolFuncRef;
+}
```
ruvi-d, line 176: You can make this one function by adding a Type param. We have functions to convert Type to string and Type to LLVMType, so no need to duplicate the code.

ruvi-d (changes requested):
Suggested some improvements

compiler/include/TypeCastInsn.h
KavinduZoysa: Changed

runtime/src/lib.rs
KavinduZoysa: Added

compiler/codegen/TypeCastInsn.cpp
KavinduZoysa: Changed

compiler/codegen/TypeCastInsn.cpp
KavinduZoysa: Changed

compiler/codegen/TypeCastInsn.cpp
KavinduZoysa: In here https://github.com/ballerina-platform/nballerina/pull/191/files#diff-6b9f0def66334b86f004fd973fd9137ff1d79244d97526871c61ffc273326f5eR164, I am checking the values that support box for now, others cases work as previous (bit casting), I think we do not need to throw and error. Anyway this line has changed to return `LLVMValueRef()`

test/casting/anyTypeReturn.bal
KavinduZoysa: We have to fix this issue https://github.com/ballerina-platform/nballerina/issues/192 for that.

compiler/include/TypeCastInsn.h
ruvi-d: `isSameType()` and `isBoxValueSupport()` are still public.
Please move `typeStringMangleName()` also

compiler/codegen/TypeCastInsn.cpp
```diff
@@ -117,13 +117,61 @@ void TypeCastInsn::translate([[maybe_unused]] LLVMModuleRef &modRef) {
...
+LLVMValueRef TypeCastInsn::getBoxValueFunc(LLVMModuleRef &modRef, LLVMTypeRef paramTypeRef, TypeTag typeTag) {
+    switch (typeTag) {
+    case TYPE_TAG_INT: {
+        return generateBoxValueFunc(modRef, paramTypeRef, "box_bal_integer");
+    }
+    case TYPE_TAG_FLOAT: {
+        return generateBoxValueFunc(modRef, paramTypeRef, "box_bal_float");
+    }
+    case TYPE_TAG_BOOLEAN: {
+        return generateBoxValueFunc(modRef, paramTypeRef, "box_bal_bool");
```
ruvi-d, line 157: Can we generate these function programmatically like :
```
std::string functionName = "box_bal_";
functionName += Type::getNameOfType(typeTag);
```

ruvi-d (changes requested):

compiler/codegen/TypeCastInsn.cpp
ruvi-d: This will be a potential pitfall when someone tries to extend the functionality in the future.
I think something like a `llvm_unreachable()` would make more sense here.

compiler/include/TypeCastInsn.h
KavinduZoysa: Fixed

compiler/codegen/TypeCastInsn.cpp
KavinduZoysa: Removed this function

compiler/codegen/TypeCastInsn.cpp
KavinduZoysa: Yes, changed

ruvi-d (approved):

runtime/src/lib.rs
```diff
@@ -187,3 +187,48 @@ pub extern "C" fn map_store_int(ptr: *mut BalMapInt, key: *const c_char, member_
...
     println!("length={}", bal_map.length());
 }
+
+#[no_mangle]
+pub extern "C" fn box_bal_int(val: i32) -> *mut i32 {
+    Box::into_raw(Box::new(val))
+}
+
+#[no_mangle]
+pub extern "C" fn deint_box_bal_int(ptr: *mut i32) {
```
manuranga, line 197: wouldn't `unbox` be more meaningful

KavinduZoysa: +1, changed

compiler/codegen/TypeCastInsn.cpp
```diff
@@ -117,13 +117,47 @@ void TypeCastInsn::translate([[maybe_unused]] LLVMModuleRef &modRef) {
...
+        LLVMBuildStore(builder, bitCastRes, elePtr2);
     } else {
         LLVMBuildBitCast(builder, rhsOpRef, lhsTypeRef, "data_cast");
     }
 }
 
+LLVMValueRef TypeCastInsn::generateBoxValueFunc(LLVMModuleRef &modRef, LLVMTypeRef paramTypeRef, TypeTag typeTag) {
+    std::string functionName = "box_bal_";
+    functionName += Type::getNameOfType(typeTag);
+    LLVMValueRef boxBoolFuncRef = getPackage()->getFunctionRef(functionName);
```
manuranga, line 138: `Bool` here is not valid anymore.

KavinduZoysa: Fixed

manuranga (approved):

Issue #157, opened 2021-03-08
Codegen for Ballerina Float is using float, not double


ruvi-d:
**Description:**

Ballerina Float type spec calls for double precision, but codegen currently uses LLVM float type which is single precision. 
https://github.com/ballerina-platform/nballerina/blob/master/compiler/llvmConstantLoadInsn.cpp#L51

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #196, opened 2021-03-23
Add support for reading float values from BIR

### LINKED ISSUE #192: Incorrect output from float value
**Description:**
$subject. 

**Steps to reproduce:**
```
public function printf32(float val) = external;


public function main() {
    printf32(2.2); // prints wrong value like 4612136000000000000
}
```

shubhamnarlawar:
Resolves #192


shubhamnarlawar:
The testcase presents in this PR expects output of `2.2` - 

1. I have created a new function for reading 8 float bytes from the BIR - `readS8bef()`
2. The hex value present in BIR is `40 01 99 99 99 99 99 9a` which is equivalent to `2.2`.
3. After reading 8 bytes from the BIR, I am getting value as - `2.2000000000000002`

I am debugging on why I am getting such value instead of `2.2`. Please take a quick look to point if there is silly mistake in reading 8 float bytes from BIR.

KavinduZoysa:
> The testcase presents in this PR expects output of `2.2` -
> 
> 1. I have created a new function for reading 8 float bytes from the BIR - `readS8bef()`
> 2. The hex value present in BIR is `40 01 99 99 99 99 99 9a` which is equivalent to `2.2`.
> 3. After reading 8 bytes from the BIR, I am getting value as - `2.2000000000000002`
> 
> I am debugging on why I am getting such value instead of `2.2`. Please take a quick look to point if there is silly mistake in reading 8 float bytes from BIR.

Cannot we get the hex value read from BIR?

ruvi-d:
> The testcase presents in this PR expects output of `2.2` -
> 
> 1. I have created a new function for reading 8 float bytes from the BIR - `readS8bef()`
> 2. The hex value present in BIR is `40 01 99 99 99 99 99 9a` which is equivalent to `2.2`.
> 3. After reading 8 bytes from the BIR, I am getting value as - `2.2000000000000002`
> 
> I am debugging on why I am getting such value instead of `2.2`. Please take a quick look to point if there is silly mistake in reading 8 float bytes from BIR.

Are you implying that `2.2000....2` is wrong and not the expected value? 
It is actually correct. Fundamentals of floating point representation in binary systems. You can't compare float values like int values. Please read up on it :-)

ruvi-d:
> Please take a quick look to point if there is silly mistake in reading 8 float bytes from BIR.

Found one.. you have forgotten to string print "RESULT="

shubhamnarlawar:
> > The testcase presents in this PR expects output of `2.2` -
> > 
> > 1. I have created a new function for reading 8 float bytes from the BIR - `readS8bef()`
> > 2. The hex value present in BIR is `40 01 99 99 99 99 99 9a` which is equivalent to `2.2`.
> > 3. After reading 8 bytes from the BIR, I am getting value as - `2.2000000000000002`
> > 
> > I am debugging on why I am getting such value instead of `2.2`. Please take a quick look to point if there is silly mistake in reading 8 float bytes from BIR.
> 
> Are you implying that `2.2000....2` is wrong and not the expected value?
> It is actually correct. Fundamentals of floating point representation in binary systems. You can't compare float values like int values. Please read up on it :-)
> 
> https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

Yes Ruvi. I expected it close to `2.2`. I tried with `4.2` and getting `4.2000...2`. So, we will consider float values as approximate values.

Thanks for the post.

shubhamnarlawar:
> > Please take a quick look to point if there is silly mistake in reading 8 float bytes from BIR.
> 
> Found one.. you have forgotten to string print "RESULT="

Thanks Ruvi for pointing it out the reason for failing the testcase.
However, my concern has been solved - whether to get exact float value or approximate float value. 

I will push the changes.

shubhamnarlawar:
> > The testcase presents in this PR expects output of `2.2` -
> > 
> > 1. I have created a new function for reading 8 float bytes from the BIR - `readS8bef()`
> > 2. The hex value present in BIR is `40 01 99 99 99 99 99 9a` which is equivalent to `2.2`.
> > 3. After reading 8 bytes from the BIR, I am getting value as - `2.2000000000000002`
> > 
> > I am debugging on why I am getting such value instead of `2.2`. Please take a quick look to point if there is silly mistake in reading 8 float bytes from BIR.
> 
> Cannot we get the hex value read from BIR?

Yes. We are reading hex values from the BIR using `BIRReader`

shubhamnarlawar:
1. Added support for printing float values.
2. Enabled float related testcases.

Please review and suggest changes if any.

KavinduZoysa:
Since we are supporting any and union, Shall we add a couple of test cases for them?

ruvi-d:
Shall we handle #157 also with this PR? You are anyway updating the Reader values from float to double so, we just need to update the Codegen classes accordingly.

compiler/BIRReader.cpp
```diff
@@ -122,6 +122,35 @@ uint64_t BIRReader::readS8be() {
     return result;
 }
 
+// Read 8 bytes from the stream for float value
+double BIRReader::readS8bef() {
+    double value;
+    double result;
+    is.read(reinterpret_cast<char *>(&value), sizeof(value));
+#if __BYTE_ORDER == __LITTLE_ENDIAN
```
ruvi-d, line 130: I believe that this is a GCC only preprocessor def. Please check whether it should work on Windows and Mac. Otherwise we will have to use a runtime check.

compiler/BIRReader.cpp
```diff
@@ -122,6 +122,35 @@ uint64_t BIRReader::readS8be() {
     return result;
 }
 
+// Read 8 bytes from the stream for float value
+double BIRReader::readS8bef() {
+    double value;
+    double result;
+    is.read(reinterpret_cast<char *>(&value), sizeof(value));
```
ruvi-d, line 129: Will size of double be 8 bytes on all platforms and architectures? Because we must read 8 bytes regardless.

compiler/BIRReader.cpp
shubhamnarlawar: You are right.

http://www.c-faq.com/cpp/ifendian.html

Our linux server is little endian machine and hence we had added the check. 
Can you tell the endianness of our platform @ruvi-d and @KavinduZoysa - as you are using both mac and windows machine.

I will check whether this flag works on little endian windows and mac machines.

compiler/BIRReader.cpp
shubhamnarlawar: @ayonam - Can you give your suggestion here please?

compiler/BIRReader.cpp
ruvi-d: We both are on Intel / Ubuntu so it's also little endian. I don't think there are any modern platforms that are using big endian. So, ideally, we just need a compiler agnostic way to check at compile time whether it's little endian (just to double check); I think it's ok to fail if otherwise.

compiler/BIRReader.cpp
shubhamnarlawar: I think irrespective of the platform, we will read 8 bytes as double value anyways. It is designed as per the BIR dump.

shubhamnarlawar: Do you mean to pass hardcoded `8` here ?
```
is.read(reinterpret_cast<char *>(&value), 8);
```

shubhamnarlawar:
> Shall we handle #157 also with this PR? You are anyway updating the BIRReader's return values from float to double so, we just need to update the Codegen classes accordingly.

Sure. I will send these changes.

compiler/BIRReader.cpp
ruvi-d: Still it won't make a difference on a system where double is not 64bit.
If we have some guarantee/reference that modern platforms are always using 64bits for doubles then we are good. Not need to explicitly handle it then.

shubhamnarlawar: Got it Ruvi.

shubhamnarlawar:
@ruvi-d - Do I need to change this line from `float` to `double`?
https://github.com/ballerina-platform/nballerina/blob/a969374bf4ee1f76fc4a33de052c35646010b2e2/compiler/codegen/Types.cpp#L44

ruvi-d:
> @ruvi-d - Do I need to change this line from `float` to `double`?
> https://github.com/ballerina-platform/nballerina/blob/a969374bf4ee1f76fc4a33de052c35646010b2e2/compiler/codegen/Types.cpp#L44

No. I don't think so. That is used to find the Runtime function names.

compiler/BIRReader.cpp
shubhamnarlawar: Done.I have added a compile time check to know the endianness of the machine.

shubhamnarlawar:
> Shall we handle #157 also with this PR? You are anyway updating the BIRReader's return values from float to double so, we just need to update the Codegen classes accordingly.

Done!

shubhamnarlawar:
1. I have fixed printing of float issue in BIRReader.
2. I have moved to `double` from `float` in codegen phase.
3. I have added a compile time machine endianness check.

@ruvi-d - Is this PR ready to merge?

compiler/BIRReader.cpp
```diff
@@ -48,6 +48,12 @@ ReadArrayStoreInsn ReadArrayStoreInsn::readArrayStoreInsn;
 ReadArrayLoadInsn ReadArrayLoadInsn::readArrayLoadInsn;
 ReadMapStoreInsn ReadMapStoreInsn::readMapStoreInsn;
 
+int BIRReader::checkEndianness() {
```
ruvi-d, line 51: Add constexpr keyword to evaluate this at compile time and not at runtime

compiler/BIRReader.cpp
```diff
@@ -69,11 +75,11 @@ uint16_t BIRReader::readS2be() {
     is.read(reinterpret_cast<char *>(&value), sizeof(value));
     char *p = (char *)&value;
     char tmp;
-#if __BYTE_ORDER == __LITTLE_ENDIAN
-    tmp = p[0];
-    p[0] = p[1];
-    p[1] = tmp;
-#endif
+    if (checkEndianness()) {
```
ruvi-d, line 78: Looks like `isLittleEndian()` is a better name for this function

compiler/codegen/ConstantLoadInsn.cpp
```diff
@@ -35,7 +35,7 @@ ConstantLoadInsn::ConstantLoadInsn(const Operand &lhs, std::shared_ptr<BasicBloc
 ConstantLoadInsn::ConstantLoadInsn(const Operand &lhs, std::shared_ptr<BasicBlock> currentBB, int intVal)
     : NonTerminatorInsn(lhs, std::move(currentBB)), typeTag(TYPE_TAG_INT), value(intVal) {}
 
-ConstantLoadInsn::ConstantLoadInsn(const Operand &lhs, std::shared_ptr<BasicBlock> currentBB, float floatVal)
+ConstantLoadInsn::ConstantLoadInsn(const Operand &lhs, std::shared_ptr<BasicBlock> currentBB, double floatVal)
```
ruvi-d, line 38: line 62 will have to get updated to use LLVMDoubleType()

ruvi-d (changes requested):
The following also needs to be updated/added:
1. Package::getLLVMTypeOfType() for Float
2. Test case for Bal float value that precision larger than 32bits

compiler/BIRReader.cpp
shubhamnarlawar: Done.

compiler/BIRReader.cpp
shubhamnarlawar: Done.

compiler/codegen/ConstantLoadInsn.cpp
shubhamnarlawar: I have changed from `LLVMFloatType` to `LLVMDoubleType()` in below cases - 
1. `Package::getLLVMTypeOfType()`
2. `ConstLoadInsn.cpp`

I am getting regression in one testcase - `./test/casting/anyTypeReturn.bal`
I am getting `32775` instead of `222`. I am debugging this issue.

shubhamnarlawar:
> The following also needs to be updated/added:
> 
>     1. Package::getLLVMTypeOfType() for Float
> 
>     2. Test case for Bal float value that precision larger than 32bits

2. Do you mean to add negative testcase like below? - 

IEEE 754 32-bit base-2 floating-point variable has a maximum value of (2 − 2−23) × 2127 ≈ 3.4028235 × 1038

assign a value greater than `3.4028235 × 10381`?

ruvi-d:
> > The following also needs to be updated/added:
> > ```
> > 1. Package::getLLVMTypeOfType() for Float
> > 
> > 2. Test case for Bal float value that precision larger than 32bits
> > ```
> 
> 1. Do you mean to add negative testcase like below? -
> 
> IEEE 754 32-bit base-2 floating-point variable has a maximum value of (2 − 2−23) × 2127 ≈ 3.4028235 × 1038
> 
> assign a value greater than `3.4028235 × 10381`?

Yes. Check the limits here: https://docs.microsoft.com/en-us/cpp/c-language/type-float?view=msvc-160#range-of-floating-point-types

shubhamnarlawar:
@ruvi-d - Is there any other change I have to make apart from - 

Changing from LLVMFloatType to LLVMDoubleType() in below cases -
1. Package::getLLVMTypeOfType()
 2. ConstLoadInsn.cpp

This is the only change I have made and one test from master is failing - `anyTypeReturn.bal` which involves 3 times casting include cast from `int` to `float` at the end.

compiler/codegen/ConstantLoadInsn.cpp
KavinduZoysa: @shubhamnarlawar, please check the rust side, because we are boxing the float value.

KavinduZoysa:
```
#[no_mangle]
pub extern "C" fn box_bal_float(val: f64) -> *mut f64 {
    Box::into_raw(Box::new(val))
}
```

KavinduZoysa: This particular issue is happening only for function call right?

ruvi-d:
> @ruvi-d - Is there any other change I have to make apart from -
> 
> Changing from LLVMFloatType to LLVMDoubleType() in below cases -
> 
> 1. Package::getLLVMTypeOfType()
> 2. ConstLoadInsn.cpp
> 
> This is the only change I have made and one test from master is failing - `anyTypeReturn.bal` which involves 3 times casting include cast from `int` to `float` at the end.
> 
> I am not clear what other change I have to make apart from suggested change.

Grep for `LLVMInt32Type`. I think there should be a few usages in the Array and TypeCast classes. Where possible use getLLVMTypeOfType() instead of hard cording the LLVMType.

shubhamnarlawar:
I nailed down the root cause of the failing testcase - 

Upon changing from `float` to `double`, a typecast from `float` to `any` is causing the testcase to fail.

shubhamnarlawar:
> I nailed down the root cause of the failing testcase -
> 
> Upon changing from `float` to `double`, a typecast from `float` to `any` is causing the testcase to fail.

I compared the `.ll` file of the failing testcase - with this branch and `master` branch. The only difference is instead of `float`, `double` is present and alignment is `8` instead of `4`.

shubhamnarlawar:
> > I nailed down the root cause of the failing testcase -
> > Upon changing from `float` to `double`, a typecast from `float` to `any` is causing the testcase to fail.
> 
> I compared the `.ll` file of the failing testcase - with this branch and `master` branch. The only difference is instead of `float`, `double` is present and alignment is `8` instead of `4`.

The testcase contains a cast as below - 

`float`->`any`->`float`->`int`

Upon casting from `float`->`int` I am getting different output every time I run the final binary - for example -
```
RESULT=32593

RESULT=32561

RESULT=32595

RESULT=32690
```

Here is the root cause of the problem in failing testcase - (snippet below)

```
function fooFloat() returns any {
    return 0.0;
}

public function main() {
    float f = <float>fooFloat();
    printu64(1 + <int>f);
}
```
Expected value - `1`
Current value - `32767`, `32765`, etc

compiler/codegen/ConstantLoadInsn.cpp
shubhamnarlawar: With this cast - `float->any->float->int`

shubhamnarlawar:
@ayonam - Can you please give your suggestions?

From my observations, at rust side for integer, we are using i32. Now we are representing float as double(64 bits).

Now, a typecast from double to int seems to cause a overflow?

So, do we have to first model bal integers as `Int64` type? Please correct me if I misunderstood.

shubhamnarlawar:
Created an issue  #221 

Now all tests are passing. @ruvi-d - Is this PR good to merge now?

shubhamnarlawar:
> Created an issue #221
> 
> Now all tests are passing. @ruvi-d - Is this PR good to merge now?

Just a moment. Master got forwarded with latest merge.

shubhamnarlawar:
> > Created an issue #221
> > Now all tests are passing. @ruvi-d - Is this PR good to merge now?
> 
> Just a moment. Master got forwarded with latest merge.

I have resolved the merge conflict.
@ruvi-d - Is this PR ready to merge now?

I have created an issue for the `float`->`int` typecast issue - #221

ruvi-d:
@shubhamnarlawar didn't you add the double precision tests that we talked about? Using a value that exceeds 32bit float point precision.

Also please link that #157 is resolved with this PR

shubhamnarlawar:
This PR resolves #157

shubhamnarlawar:
> @shubhamnarlawar didn't you add the double precision tests that we talked about? Using a value that exceeds 32bit float point precision.
> 
> Also please link that #157 is resolved with this PR

I tried below testcase by assigning value `3.402823466 E + 38` i.e. `340282346600000064440896728032664552640`

```
public function main() {
    float f = 340282346600000064440896728032664552640;
    print_string("RESULT=");
    printf64(f);
}
```

But I am getting jballerina error as below - 

`error: .::doublePrecisionLimit.bal:8:15: Integer '340282346600000064440896728032664552640' too large`

ruvi-d:
> This PR resolves #157

You have to put it in the PR description

ruvi-d:
> > @shubhamnarlawar didn't you add the double precision tests that we talked about? Using a value that exceeds 32bit float point precision.
> > Also please link that #157 is resolved with this PR
> 
> I tried below testcase by assigning value `3.402823466 E + 38` i.e. `340282346600000064440896728032664552640`
> 
> ```
> public function main() {
>     float f = 340282346600000064440896728032664552640;
>     print_string("RESULT=");
>     printf64(f);
> }
> ```
> 
> But I am getting jballerina error as below -
> 
> `error: .::doublePrecisionLimit.bal:8:15: Integer '340282346600000064440896728032664552640' too large`

That's an int literal isn't it? 
you just need to find a float literal that's precision is between 32bit and 64bit.

shubhamnarlawar:
> > > @shubhamnarlawar didn't you add the double precision tests that we talked about? Using a value that exceeds 32bit float point precision.
> > > Also please link that #157 is resolved with this PR
> > 
> > 
> > I tried below testcase by assigning value `3.402823466 E + 38` i.e. `340282346600000064440896728032664552640`
> > ```
> > public function main() {
> >     float f = 340282346600000064440896728032664552640;
> >     print_string("RESULT=");
> >     printf64(f);
> > }
> > ```
> > 
> > 
> > But I am getting jballerina error as below -
> > `error: .::doublePrecisionLimit.bal:8:15: Integer '340282346600000064440896728032664552640' too large`
> 
> That's an int literal isn't it?
> you just need to find a float literal that's precision is between 32bit and 64bit.

Yes Ruvi. I meant to put a `.` like `3.4` and yes value > 32 bit is working.

With new jballerina pack - I am getting an error - 
```
/home/shubham/float/nballerina/compiler/BIRReader.cpp:582 Invalid Insn Kind for Reader.
Segmentation fault (core dumped)
```

I am checking on why I am getting this kind of error. With old pack, it is working fine. Can you please tell the difference between old and new jballerina pack.

ruvi-d:
> I am checking on why I am getting this kind of error. With old pack, it is working fine. Can you please tell the difference between old and new jballerina pack.

Replied here: https://ballerina-platform.slack.com/archives/C01LNSAHK18/p1617025891112800

shubhamnarlawar:
> You have to

Done.

shubhamnarlawar:
@ruvi-d - Please check the testcase named - `floatValueUpperLimit.bal` for checking the upper limit of float value. 

I have addressed all the suggestions from this PR.

compiler/codegen/Package.cpp
```diff
@@ -72,7 +72,7 @@ LLVMTypeRef Package::getLLVMTypeOfType(const Type &type) const {
         return LLVMInt32Type();
     case TYPE_TAG_BYTE:
     case TYPE_TAG_FLOAT:
-        return LLVMFloatType();
```
ruvi-d, line 75: `case TYPE_TAG_BYTE:` is returning `LLVMFloatType()`

test/casting/anyTypeReturn.bal
```diff
@@ -14,19 +12,14 @@ function fooBool() returns any {
...
 public function main() {
     print_string("RESULT=");
     int i = <int>fooInt();
     boolean b = <boolean>fooBool();
     if (b) {
         i = i + 1;
     }
-    float f = <float>fooFloat();
-    printu32(i + <int>f);
+    printu32(i);
```
ruvi-d, line 22: Since this test is for any type return, I think we need to cover all the basic types. Add a string return too and put a TODO for float

test/casting/floatTypecast.bal
```diff
@@ -9,6 +13,8 @@ public function bar(any z) returns float
 public function main() {
     float b = 10.5;
     float c = bar(b);
+    print_string("RESULT=");
+    printf64(c);
```
ruvi-d, line 17: for float-any type casting just keep everything in main. No need to also cover any function param test here.

test/casting/anyTypeReturn.bal
```diff
@@ -4,8 +4,6 @@ public function print_string(string val) = external;
 
 public function printu32(int val) = external;
 
-public function printf32(float val) = external;
```
ruvi-d, line 7: test file name needs to be snake case

test/casting/floatTypecast.bal
```diff
@@ -1,5 +1,9 @@
 // RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
 
+public function print_string(string val) = external;
```
ruvi-d, line 3: test file name needs to be snake case

test/float-type/floatUpperLimit.bal
```diff
@@ -0,0 +1,14 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
+
+public function print_string(string val) = external;
```
ruvi-d, line 3: test file name needs to be snake case

test/float-type/floatUpperLimit.bal
```diff
@@ -0,0 +1,14 @@
...
+
+public function printf64(float val) = external;
+
+public function main() {
+    print_string("RESULT=");
+    //Passing value greater than upper limit of float i.e. 3.402823466 E + 39
+    float f = 3.402823465999999976608952504680408952584;
```
ruvi-d, line 10: i think a better test would be: (ballerina syntax)
```
import ballerina/io;
public function main() {
    float val = 22.0/7.0;
    io:println(val);
}
```
Expected value for double precision is : `3.142857142857143`
Single precision output would be some thing like : `3.142857074737549`

I used https://play.ballerina.io/ and the following code to verify
```
#include <stdio.h>

int main(){
    float val1 = 22.0f/7.0f;
    printf("%.15f\n", val1);
    double val2 = 22.0l/7.0l;
    printf("%.15f\n", val2);
}
```

ruvi-d (changes requested):

ruvi-d:
missing tests:
* Global float
* Separate test for float function args (with multiple args too)
* Negative float value init
* Float arithmetic (addition, subtraction,  multiplication, division, unary ops, etc)
* Float comparison ( greater or equal, etc)
* Float type function return

KavinduZoysa:
Adding to @ruvi,

- float as a member of union type
- float as any type

compiler/codegen/Package.cpp
shubhamnarlawar: It should it be `LLVMInt32Type()` right?

test/casting/anyTypeReturn.bal
shubhamnarlawar: In tests and compiler, we are following `CamelCase`.

compiler/codegen/Package.cpp
ruvi-d: I think we don't use it anymore. You should be able to remove the switch case for that

test/casting/anyTypeReturn.bal
ruvi-d: not with #213

shubhamnarlawar:
- [] Global float
- [] Separate test for float function args (with multiple args too)
      Negative float value init
* Float arithmetic (addition, subtraction,  multiplication, division, unary ops, etc)
* Float comparison ( greater or equal, etc)
* Float type function return

compiler/BIRReader.cpp
```diff
@@ -48,6 +48,12 @@ ReadArrayStoreInsn ReadArrayStoreInsn::readArrayStoreInsn;
 ReadArrayLoadInsn ReadArrayLoadInsn::readArrayLoadInsn;
 ReadMapStoreInsn ReadMapStoreInsn::readMapStoreInsn;
 
+constexpr int BIRReader::isLittleEndian() {
```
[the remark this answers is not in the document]
shubhamnarlawar: Sure. Updated to bool return.

test/casting/floatTypecast.bal
shubhamnarlawar: Done.

test/float-type/floatUpperLimit.bal
shubhamnarlawar: Done.

test/float-type/floatUpperLimit.bal
shubhamnarlawar: Division on float require  generation of new instruction rather than div. Shall I create an issue for support for floating point arithmetics? 

Let us just pass this PR to fix incorrect float value bug.

shubhamnarlawar: Same goes for comparision operation. For `a>b`, I am getting different instruction as `icmp ult`.

shubhamnarlawar:
> Adding to @ruvi,
> 
>     * float as a member of union type
> 
>     * float as any type

I have added these two testcase!

shubhamnarlawar:
@ruvi-d - For global float testcase, currently we are ignoring `init()` from BIR. I checked `./test/any/cast_global_any_int.bal` - 

```
any a = 23;

public function main() {
    a = 23;    // TODO remove ressaignment when global init is fixed
    int x = <int>a;
    print_string("RESULT=");
    printu32(x);
}
```

Do you want me to add global testcase as above?

ruvi-d:
> @ruvi-d - For global float testcase, currently we are ignoring `init()` from BIR. I checked `./test/any/cast_global_any_int.bal` - 
> 
> ```
> any a = 23;
> 
> public function main() {
>     a = 23;    // TODO remove ressaignment when global init is fixed
>     int x = <int>a;
>     print_string("RESULT=");
>     printu32(x);
> }
> ```
> 
> Do you want me to add global testcase as above?

Yes. Just reassign and use.

test/float-type/floatUpperLimit.bal
shubhamnarlawar: Created an issue for above - #241

shubhamnarlawar:
> // TODO remove ressaignment when global init is fixed

Done.

test/casting/floatTypecast.bal
shubhamnarlawar: Done.

shubhamnarlawar:
@ruvi-d - Please review and suggest changes.
Now, the PR contains fix for incorrect float value reading and testcases related to it.

test/float/function_return.bal
```diff
@@ -0,0 +1,25 @@
...
+
+public function foo(float a) returns float
+{
+    return a;
+}
+
+public function bar(float a, float b) returns float
+{
+    return b;
+}
```
ruvi-d, line 15: I don't think we need two functions. just one function call should be enough to cover this test case

test/float/float.bal
```diff
@@ -0,0 +1,12 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
+
+public function print_string(string val) = external;
+
+public function printf32(float val) = external;
+
+public function main() {
+    print_string("RESULT=");
```
ruvi-d, line 8: Add a float type named variable and print also

test/float/float.bal
```diff
@@ -0,0 +1,12 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
```
ruvi-d, line 1: better file name would be `init_float.bal` ?

test/float/global_float.bal
```diff
@@ -0,0 +1,15 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
```
ruvi-d, line 1: better file name would be `init_float_global.bal` ?

ruvi-d (changes requested):

test/float/cast_any_float.bal
```diff
@@ -0,0 +1,20 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
```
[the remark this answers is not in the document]
shubhamnarlawar: So, its a duplicate testcase. Should I remove it?

test/float/float.bal
```diff
@@ -0,0 +1,12 @@
...
+
+public function print_string(string val) = external;
+
+public function printf32(float val) = external;
+
+public function main() {
+    print_string("RESULT=");
+    printf32(2.2);
```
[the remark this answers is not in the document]
shubhamnarlawar: Negative value support is in `init_float.bal`. In fact, we  don't need `float.bal`. I will remove it.

test/float/function_return.bal
shubhamnarlawar: Removed single param function.

test/float/global_float.bal
shubhamnarlawar: Done. Thanks for the suggestion!

test/float/float.bal
shubhamnarlawar: Removed `float.bal` as `init_float.bal` is already present.

test/float/float.bal
shubhamnarlawar: Removed float.bal as init_float.bal covers above functionality.

test/any/return_any_basic_types.bal
```diff
@@ -26,8 +20,6 @@ public function main() {
         i = i + 1;
     }
     // TODO check float return
-    //float f = <float>fooFloat();
-    // printu32(i + <int>f);
```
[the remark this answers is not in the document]
shubhamnarlawar: Done.

test/float/cast_any_float.bal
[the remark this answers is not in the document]
shubhamnarlawar: Removed `cast_any_float.bal` as the functionality is already covered in `/test/any/return_any_basic_types.bal`

test/any/return_any_basic_types.bal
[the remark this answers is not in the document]
ruvi-d: missed the point here. Test is to check the return functionality of all the basic types. So we need to print `// CHECK: RESULT=222` and `// CHECK: RESULT=7.1`
Also, delete the TODO comment

ruvi-d (changes requested):

test/any/return_any_basic_types.bal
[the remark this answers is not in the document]
shubhamnarlawar: Done. Thanks
Please check it.

ruvi-d (approved):

shubhamnarlawar:
Please approve this PR - @Kishanthan @uthaiyashankar @manuranga @KavinduZoysa

KavinduZoysa (approved):

PR #200, opened 2021-03-24
Support for optional type

### LINKED ISSUE #173: Support optional type
**Description:**
Optional Types
- [ ] Design - 1 PD
- [ ] BIR to LLVM IR - 2 PD
- [ ] Rust side - 2 PD
- [ ] Typecasts, Params, Return types - 2 PD

**Describe your problem(s)**

**Describe your solution(s)**

KavinduZoysa:
## Purpose
$subject
Fixes #173

### DIFF
--- a/compiler/codegen/TypeCastInsn.cpp
+++ b/compiler/codegen/TypeCastInsn.cpp
@@ -48,7 +48,7 @@ void TypeCastInsn::translate(LLVMModuleRef &modRef) {
     const auto &rhsType = rhsVar->getType();
     auto rhsTypeTag = rhsType.getTypeTag();
 
-    const char *inherentTypeChar = "inherentTypeIdx";
+    const char *inherentTypeName = "inherentTypeName";
 
     if (rhsTypeTag == TYPE_TAG_ANY || rhsTypeTag == TYPE_TAG_UNION) {
         if (lhsTypeTag == TYPE_TAG_UNION || lhsTypeTag == TYPE_TAG_ANY) {
@@ -57,7 +57,7 @@ void TypeCastInsn::translate(LLVMModuleRef &modRef) {
             return;
         }
         // GEP of last type of smart pointer(original type of any variable(smart pointer))
-        LLVMValueRef inherentTypeIdx = LLVMBuildStructGEP(builder, rhsOpRef, 0, inherentTypeChar);
+        LLVMValueRef inherentTypeIdx = LLVMBuildStructGEP(builder, rhsOpRef, 0, inherentTypeName);
         LLVMValueRef inherentTypeLoad = LLVMBuildLoad(builder, inherentTypeIdx, "");
         LLVMValueRef sExt = LLVMBuildSExt(builder, inherentTypeLoad, LLVMInt64Type(), "");
 
@@ -88,7 +88,7 @@ void TypeCastInsn::translate(LLVMModuleRef &modRef) {
     } else if (lhsTypeTag == TYPE_TAG_ANY || lhsTypeTag == TYPE_TAG_UNION) {
 
         // struct first element original type
-        LLVMValueRef inherentTypeIdx = LLVMBuildStructGEP(builder, lhsOpRef, 0, inherentTypeChar);
+        LLVMValueRef inherentTypeIdx = LLVMBuildStructGEP(builder, lhsOpRef, 0, inherentTypeName);
         std::string_view rhsTypeName = typeStringMangleName(lhsTypeRef, rhsTypeTag, rhsType);
         getPackageMutableRef().addToStrTable(rhsTypeName);
         int tempRandNum1 = rand() % 1000 + 1;
--- a/compiler/codegen/TypeCastInsn.cpp
+++ b/compiler/codegen/TypeCastInsn.cpp
@@ -172,6 +172,9 @@ std::string_view TypeCastInsn::typeStringMangleName(LLVMTypeRef valType, TypeTag
     case TYPE_TAG_ANY: {
         return "__X";
     }
+    case TYPE_TAG_NIL: {
+        return "__N";
+    }
     default:
         return "";
     }
--- a/compiler/codegen/TypeCastInsn.cpp
+++ b/compiler/codegen/TypeCastInsn.cpp
@@ -43,22 +43,23 @@ void TypeCastInsn::translate(LLVMModuleRef &modRef) {
     assert(lhsVar.has_value());
     auto rhsVar = funcObj.getLocalOrGlobalVariable(rhsOp);
     assert(rhsVar.has_value());
-    auto lhsType = lhsVar->getType().getTypeTag();
-    auto rhsType = rhsVar->getType().getTypeTag();
+    const auto &lhsType = lhsVar->getType();
+    auto lhsTypeTag = lhsType.getTypeTag();
+    const auto &rhsType = rhsVar->getType();
+    auto rhsTypeTag = rhsType.getTypeTag();
 
-    const char *lastTypeChar = "lastTypeIdx";
     const char *inherentTypeChar = "inherentTypeIdx";
 
-    if (rhsType == TYPE_TAG_ANY || rhsType == TYPE_TAG_UNION) {
-        if (lhsType == TYPE_TAG_UNION || lhsType == TYPE_TAG_ANY) {
+    if (rhsTypeTag == TYPE_TAG_ANY || rhsTypeTag == TYPE_TAG_UNION) {
+        if (lhsTypeTag == TYPE_TAG_UNION || lhsTypeTag == TYPE_TAG_ANY) {
             LLVMValueRef rhsVarOpRef = funcObj.createTempVariable(rhsOp);
             LLVMBuildStore(builder, rhsVarOpRef, lhsOpRef);
             return;
         }
         // GEP of last type of smart pointer(original type of any variable(smart pointer))
-        LLVMValueRef lastTypeIdx = LLVMBuildStructGEP(builder, rhsOpRef, 0, lastTypeChar);
-        LLVMValueRef lastTypeLoad = LLVMBuildLoad(builder, lastTypeIdx, "");
-        LLVMValueRef sExt = LLVMBuildSExt(builder, lastTypeLoad, LLVMInt64Type(), "");
+        LLVMValueRef inherentTypeIdx = LLVMBuildStructGEP(builder, rhsOpRef, 0, inherentTypeChar);
+        LLVMValueRef inherentTypeLoad = LLVMBuildLoad(builder, inherentTypeIdx, "");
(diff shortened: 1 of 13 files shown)


ruvi-d:
Shall we add a test case for global optional variables too?

KavinduZoysa:
> And what about optional complex types like arrays and maps?

What you mean here is, `map<int>? marksOrNil = {};` or `map<int?> marks = {};`?

ruvi-d:
Only `map<int>?` at the moment because we only have support for map's of int type

runtime/src/lib.rs
```diff
@@ -39,8 +39,12 @@ pub extern "C" fn is_same_type(src_type: *const c_char, dest_type: *const c_char
...
     let dest_cstr: &CStr = unsafe { CStr::from_ptr(dest_type) };
     //Conversion to rust strings
-    let source: String = source_cstr.to_str().unwrap().to_owned();
-    let destination: String = dest_cstr.to_str().unwrap().to_owned();
+    let mut source: String = source_cstr.to_str().unwrap().to_owned();
+    let mut destination: String = dest_cstr.to_str().unwrap().to_owned();
+    source.remove(0);
+    source.remove(0);
+    destination.remove(0);
+    destination.remove(0);
```
ruvi-d, line 47: remove looks like a very expensive operation to run twice like this on the same string. 
I think you can use string slices and ranges for this like: `let filtered_source = &source[2..];` 
Please check

ruvi-d (changes requested):

runtime/src/lib.rs
ruvi-d: i suppose that we should also do a length check

ruvi-d (approved):

test/optional_type/basic_optional_types3.bal
```diff
@@ -0,0 +1,14 @@
...
+
+public function print_str(string val) = external;
+
+public function printu32(int val) = external;
+
+public function main() {
+    string? x = ();
+    () n = <()>x;
+    print_str("RESULT=");
+    printu32(231);
```
ayonam, line 11: This test doesn't make much sense other than ascertaining that the compiler is generating some code or not.  How correct is the code is not reflected by this test.  So the value 'n' needs to have some kind of check to ensure that the computation we performed, went as per expectation.

test/optional_type/optional_type_in_array.bal
```diff
@@ -0,0 +1,14 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
+
+public function print_str(string val) = external;
+
+public function printu32(int val) = external;
+
+public function main() {
+    int[]? arrOrNil = [];
```
ayonam, line 8: We should have an initializer here for arrOrNil such that arrOrNil[0] has a value other than 32 to start with.  Thereafter the assignment to arr[0] should overwrite that and we should check if we got the new value or not.

test/optional_type/optional_type_in_func_arg1.bal
```diff
@@ -0,0 +1,15 @@
...
+
+public function printu32(int val) = external;
+
+function foo(int? i) returns int {
+    return <int>i;
+}
+
+public function main() {
+    print_str("RESULT=");
+    printu32(foo(32));
```
ayonam, line 13: There should also be a case where we pass a NIL to foo and check whether we print as expected.

test/optional_type/optional_type_in_func_arg2.bal
```diff
@@ -0,0 +1,13 @@
...
+
+public function print_str(string val) = external;
+
+function foo(string? s) returns string {
+    return <string>s;
+}
+
+public function main() {
+    print_str("RESULT=");
+    print_str(foo("Test"));
```
ayonam, line 11: Similar check here for a NIL being passed to foo().

test/optional_type/optional_type_in_global_var.bal
```diff
@@ -0,0 +1,14 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
+
+public function print_str(string val) = external;
+
+public function printu32(int val) = external;
+
+int? a = 24;
+
+public function main() {
+    a = 24;
```
ayonam, line 10: This assignment should be a value other than 24.  Then we should see if we are actually assigning to the right location and the old value is being overwritten or not.

test/optional_type/optional_type_in_map.bal
```diff
@@ -0,0 +1,8 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
+
+public function main() {
+    map<int>? marksOrNil = {};
+    map<int> marks = <map<int>>marksOrNil;
+    marks["jon"] = 60;
```
ayonam, line 6: Again, this test should start with an assigned value in marksOrNil.  Then assign a different value to marks["jon"] and see if the assignment went through properly.

test/optional_type/optional_type_in_return.bal
```diff
@@ -0,0 +1,16 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" | filecheck %s
+
+public function print_str(string val) = external;
+
+public function printu32(int val) = external;
+
+function foo(int i) returns int? {
+    return i;
```
ayonam, line 8: You could return i + 1 to ensure that the call is not optimized out and the passed value printed.

ayonam (changes requested):

runtime/src/lib.rs
```diff
@@ -41,7 +41,14 @@ pub extern "C" fn is_same_type(src_type: *const c_char, dest_type: *const c_char
     //Conversion to rust strings
     let source: String = source_cstr.to_str().unwrap().to_owned();
     let destination: String = dest_cstr.to_str().unwrap().to_owned();
-    return type_checker::same_type(source, destination);
+
+    if (source.len() < 2) || (destination.len() < 2) {
```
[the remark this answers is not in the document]
ruvi-d: IMO spurious string checks shouldn't be part of a release build. It's not possible unless there is a bug in the LLVM codegen. How about something like a debug build only `assert` (like we do in C++) for both underscores and length checks?

Regarding the length value. I think `<2` is correct since length of 3 is a valid string and 2 isn't.

test/optional_type/optional_type_in_return.bal
KavinduZoysa: Fixed

test/optional_type/optional_type_in_map.bal
KavinduZoysa: Fixed

test/optional_type/optional_type_in_global_var.bal
KavinduZoysa: Fixed

test/optional_type/optional_type_in_array.bal
KavinduZoysa: Fixed

compiler/codegen/TypeCastInsn.cpp
```diff
@@ -43,22 +43,23 @@ void TypeCastInsn::translate(LLVMModuleRef &modRef) {
...
+        if (lhsTypeTag == TYPE_TAG_UNION || lhsTypeTag == TYPE_TAG_ANY) {
             LLVMValueRef rhsVarOpRef = funcObj.createTempVariable(rhsOp);
             LLVMBuildStore(builder, rhsVarOpRef, lhsOpRef);
             return;
         }
         // GEP of last type of smart pointer(original type of any variable(smart pointer))
-        LLVMValueRef lastTypeIdx = LLVMBuildStructGEP(builder, rhsOpRef, 0, lastTypeChar);
-        LLVMValueRef lastTypeLoad = LLVMBuildLoad(builder, lastTypeIdx, "");
-        LLVMValueRef sExt = LLVMBuildSExt(builder, lastTypeLoad, LLVMInt64Type(), "");
+        LLVMValueRef inherentTypeIdx = LLVMBuildStructGEP(builder, rhsOpRef, 0, inherentTypeChar);
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed

test/optional_type/basic_optional_types3.bal
KavinduZoysa: @ayonam, I agree with you. But here we have a limitation that still we did not implement the `is-expr`(https://github.com/ballerina-platform/nballerina/issues/220). Ideally, we need to do a type check here, like `x is ()`. 

Also, we cannot do a type check like `n is ()` because it is a compile-time error. 
`unnecessary condition: expression will always evaluate to 'true'`

What we can do is assert the runtime error if `<()>x` fails, which is still a limitation.

```
    print_str("RESULT=");
    printu32(231);
```
I have a reason to add the above 2 lines. Based on the result return by runtime type checking(`is_same_type`) we split the basic block into two(split and abort). Above two lines are placed inside the `.split` basic block and if runtime type check passes `RESULT=231` should be printed. I know this is not a good approach, but with the current limitations I cannot do something better than this, until we fix https://github.com/ballerina-platform/nballerina/issues/220.

test/optional_type/optional_type_in_func_arg1.bal
KavinduZoysa: We have to wait until we fix this issue https://github.com/ballerina-platform/nballerina/issues/220 to check the nil type

test/optional_type/optional_type_in_func_arg2.bal
KavinduZoysa: We have to wait until we fix this issue https://github.com/ballerina-platform/nballerina/issues/220 to check the nil type

runtime/src/lib.rs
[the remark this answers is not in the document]
ayonam: Okay.  You can have an assert in the debug builds.  But usually, we have situations where it is llvm_unreachable() that is used.  That could be a better way.

If length of 3 is valid string then the check should be <=2 or <3 for an assertion to fail.

test/optional_type/basic_optional_types3.bal
ayonam: Yes, I agree with you.  But we should add a note in the relevant PR for the implementation of the <()>x that this test should be modified when that feature is implemented.

test/optional_type/optional_type_in_func_arg1.bal
ayonam: Agreed.

runtime/src/lib.rs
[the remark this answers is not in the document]
ayonam: @KavinduZoysa Can you please make the necessary changes for this one?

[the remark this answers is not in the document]
KavinduZoysa: @ayonam, I have done the change under this commit 75513c8, please check

ayonam (approved):

PR #235, opened 2021-04-04
Model Bal Int as 64 bits in codegen phase 

### LINKED ISSUE #156: Codegen for Ballerina Int is using 32bit, not 64bit
**Description:**

Ballerina spec specify that Ballerina Int is of type 64 bit signed integers; but codegen for Ballerina Int types currently uses 32 bit Integers: https://github.com/ballerina-platform/nballerina/blob/master/compiler/llvmConstantLoadInsn.cpp#L47

Need to also add a test to cover the integers larger that 32 bits. 

**Steps to reproduce:**

**Affected Versions:**

shubhamnarlawar:
Resolves #156


shubhamnarlawar:
I have changed from `LLVMInt32Type()` to `LLVMInt64Type()` in `Package.cpp` and `ConstLoadInsn.cpp`.

@ruvi-d - Please suggest on how to handle maps in this case.

ruvi-d:
> @ruvi-d - Please suggest on how to handle maps in this case.

Same as for array of int types. The corresponding runtime functions need to be updated.

shubhamnarlawar:
@ruvi-d - Please review and suggest changes.

ruvi-d:
Please:
* Add a test for int value that is larger than 32 bits
* Remove 32 bit print int functions from runtime lib (update the tests accordingly)

shubhamnarlawar:
> Please:
> 
>     * Add a test for int value that is larger than 32 bits
> 
>     * Remove 32 bit print int functions from runtime lib (update the tests accordingly)



> Please:
> 
>     * Add a test for int value that is larger than 32 bits
> 
>     * Remove 32 bit print int functions from runtime lib (update the tests accordingly)

1. I have added a test which has a int value `> 32 bits` - `greater_than_32_bits.bal`
2. I have replaced all `printu32()` with `printu64()` in all bal tests.
3. I am using `uint64_t` instead of `int` as `int` just hold 4 bytes. Is this fine?

ruvi-d:
> 1. I have added a test which has a int value `> 32 bits` - `greater_than_32_bits.bal`

In the same test, print both positive and negative values that are larger than 32bit

> 2. I have replaced all `printu32()` with `printu64()` in all bal tests.

We should be using only signed 64 values. i.e `print_i64` not `print_u64`. Ballerina int's are signed and 64bit

> 3. I am using `uint64_t` instead of `int` as `int` just hold 4 bytes. Is this fine?

Again, you can't use unsigned. I assume that `long long` or `int64_t` will work

test/any/cast_any_basic_types.bal
```diff
@@ -2,7 +2,7 @@
 
 public function print_string(string val) = external;
 
-public function printu32(int val) = external;
+public function printu64(int val) = external;
```
ruvi-d, line 5: Function not in use. Can be deleted.

test/int/greater_than_32_bits.bal
```diff
@@ -0,0 +1,12 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" "%skip_bir_gen" | filecheck %s
+
+public function print_string(string val) = external;
+
+public function printu64(int val) = external;
+
+public function main() {
+    int val = 8589934591;
```
ruvi-d, line 8: Need to test a negative value too

ruvi-d (changes requested):
* Need to change all u64 references to i64
* Need to add test cases to cover negative value ints
* Delete Rust print functions that we are not using

test/int/greater_than_32_bits.bal
shubhamnarlawar: Done. I have added negative value also in `greater_than_32_bits.bal`

shubhamnarlawar:
> > 1. I have added a test which has a int value `> 32 bits` - `greater_than_32_bits.bal`
> 
> In the same test, print both positive and negative values that are larger than 32bit

Done.

> 
> > 1. I have replaced all `printu32()` with `printu64()` in all bal tests.
> 
> We should be using only signed 64 values. i.e `print_i64` not `print_u64`. Ballerina int's are signed and 64bit

Got it. We have `print64`. Let me rename it to `print_i64` and update all the testcases.

> 
> > 1. I am using `uint64_t` instead of `int` as `int` just hold 4 bytes. Is this fine?
> 
> Again, you can't use unsigned. I assume that `long long` or `int64_t` will work

`int64_t` seems fine to me.

test/any/cast_any_basic_types.bal
shubhamnarlawar: Deleted all unsigned print int functions from rust.

shubhamnarlawar:
>     * Need to change all u64 references to i64

Done.

> 
>     * Need to add test cases to cover negative value ints

Done in `greater_than_32_bits.bal` and `init_int.bal`.

> 
>     * Delete Rust print functions that we are not using

Removed unsued rust functions.

shubhamnarlawar:
Is it good to merge now?

compiler/BIRReader.h
```diff
@@ -313,7 +313,7 @@ class ConstantPoolSet {
     BIRReader &readerRef = BIRReader::reader;
     ConstantPoolEntry *getEntry(int index) { return poolEntries[index].get(); }
     std::string getStringCp(uint32_t index);
-    uint32_t getIntCp(uint32_t index);
+    uint64_t getIntCp(uint32_t index);
```
ruvi-d, line 316: Still using unsigned int

runtime/src/lib.rs
```diff
@@ -86,49 +86,25 @@ pub extern "C" fn is_same_type(src_type: *const c_char, dest_type: *const c_char
...
 #[no_mangle]
-pub extern "C" fn print64(num64: i64) {
+pub extern "C" fn print_i64(num64: i64) {
     println!("{}", num64);
 }
 
 // Prints 32 bit signed integer
 #[no_mangle]
-pub extern "C" fn print32(num32: i32) {
+pub extern "C" fn print_i32(num32: i32) {
```
ruvi-d, line 95: There is no need to use this function anymore

runtime/src/lib.rs
```diff
@@ -86,49 +86,25 @@ pub extern "C" fn is_same_type(src_type: *const c_char, dest_type: *const c_char
...
     println!("{}", num32);
 }
 
 // Prints 16 bit signed integer
 #[no_mangle]
-pub extern "C" fn print16(num16: i16) {
+pub extern "C" fn print_i16(num16: i16) {
```
ruvi-d, line 101: We don't need this either right?

runtime/src/lib.rs
```diff
@@ -86,49 +86,25 @@ pub extern "C" fn is_same_type(src_type: *const c_char, dest_type: *const c_char
...
-// Prints 16 bit unsigned integer
-#[no_mangle]
-pub extern "C" fn printu16(num16: u16) {
-    println!("{}", num16);
-}
-
-// Prints 8 bit unsigned integer
-#[no_mangle]
-pub extern "C" fn printu8(num8: u8) {
+pub extern "C" fn print_i8(num8: i8) {
```
ruvi-d, line 107: Do we need print_i8?

runtime/src/lib.rs
```diff
@@ -86,49 +86,25 @@ pub extern "C" fn is_same_type(src_type: *const c_char, dest_type: *const c_char
 
 // Prints 64 bit signed integer
 #[no_mangle]
-pub extern "C" fn print64(num64: i64) {
+pub extern "C" fn print_i64(num64: i64) {
```
ruvi-d, line 89: `print_integer` would have been a more appropriate name here

ruvi-d (changes requested):

compiler/BIRReader.h
shubhamnarlawar: During initial stage, @ayonam suggested to read all the values from the BIR dump as unsigned values and not signed values.

shubhamnarlawar: All byte stream functions return unsigned values. As per your suggestion to store it as `int64_t`, I am doing casting from `uint64_t` to `int64_t`.

runtime/src/lib.rs
shubhamnarlawar: No. All functions can use `print_i64`.
Removed.

runtime/src/lib.rs
shubhamnarlawar: Removed.

runtime/src/lib.rs
shubhamnarlawar: As `print_i32`, `print_i16` and `print_i8` are removed. `print_integer` seems a good name. Thanks for the suggestion.

runtime/src/lib.rs
shubhamnarlawar: Removed.

compiler/BIRReader.h
ruvi-d: BIR spec says that it's 8 bytes and signed : https://github.com/ballerina-platform/ballerina-lang/blob/nballerina/docs/bir-spec/src/main/resources/kaitai/bir.ksy#L49

ruvi-d (changes requested):
One issue to resolve: https://github.com/ballerina-platform/nballerina/pull/235#discussion_r607951884

shubhamnarlawar:
> One issue to resolve: [#235 (comment)](https://github.com/ballerina-platform/nballerina/pull/235#discussion_r607951884)

Done.

compiler/BIRReader.h
shubhamnarlawar: Done. I have updated all the places. Thanks for the suggestion.

ruvi-d (approved):

shubhamnarlawar:
Please approve this PR - @Kishanthan @uthaiyashankar @manuranga @KavinduZoysa

shubhamnarlawar:
Merging PR #196 causes merge conflict in this PR because both PR were related to changes in all the testcases.

KavinduZoysa (approved):

Issue #237, opened 2021-04-05
Handle invalid keys access for maps


ruvi-d:
**Description:**

Continues the work from #128 
The implementation details are at #230 

**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #238, opened 2021-04-05
Add support for map<int> load

### LINKED ISSUE #128: Support map get member from key
**Description:**
Support map get member from key. Depends on `Union`(#112 ) and `nil`(#94 ) type support.
Continues the work from #89 

**Suggested Labels:**

Task/Feature

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

### LINKED ISSUE #216: Migrate runtime maps to use rust string
**Description:**

**Describe your problem(s)**

**Describe your solution(s)**

ruvi-d:
## Purpose

Add support to load valid keys only. A new issue (#237) has been created to separately handle invalid keys. 

Resolves #128 
Resolves #216


runtime/src/lib.rs
```diff
@@ -339,23 +339,56 @@ pub extern "C" fn map_deint_int(ptr: *mut BalMapInt) {
...
+    assert!(!key.is_null());
+    let key_str = unsafe { (*key).value };
+
+    // Output param
+    assert!(!output_val.is_null());
+
+    match bal_map.get(key_str) {
+        Some(val) => {
+            unsafe { *output_val = val.clone() };
+            true
```
KavinduZoysa, line 373: Just to know, why are we cloning here?

compiler/codegen/Function.cpp
```diff
@@ -178,6 +178,55 @@ void Function::patchBasicBlocks() {
     }
 }
 
+void Function::addValueToSmartStruct(LLVMModuleRef &modRef, LLVMValueRef value, const Type &valueType,
```
KavinduZoysa, line 181: I feel this function name do not explain what we are doing here. As I know, we are adding the type too, not only value. So I suggest something like `generateSmartStruct`. WDYT?

runtime/src/lib.rs
ruvi-d: Sorry I missed that. I should have deferenced instead of  making a copy. Will fix.

compiler/codegen/Function.cpp
ruvi-d: How about `storeValueInSmartStruct`? 
I see your point about `addValue`, but `generateSmartStruct` seem a bit much since we are only assigning to it.

KavinduZoysa: +1

KavinduZoysa (approved):

compiler/codegen/MapInsns.cpp
```diff
@@ -57,4 +57,42 @@ void MapStoreInsn::codeGenMapStore(LLVMBuilderRef builder, LLVMValueRef mapStore
...
+    LLVMValueRef outParam = LLVMBuildAlloca(builder, LLVMInt32Type(), "_out_param");
+    LLVMValueRef params[] = {funcObj.createTempVariable(rhsOp), funcObj.createTempVariable(keyOp), outParam};
+    
+    [[maybe_unused]] LLVMValueRef retVal = LLVMBuildCall(builder, getMapLoadDeclaration(modRef), params, 3, "");
+
+    // TODO check retVal and branch 
+    // if retVal is true
+    getFunctionMutableRef().storeValueInSmartStruct(modRef, outParam, Type(TYPE_TAG_INT, ""), lhs);
+    // else
+    // getFunctionMutableRef().storeValueInSmartStruct(modRef, getPackageRef().getGlobalNilVar(), Type(TYPE_TAG_NIL, ""), lhs);
```
Kishanthan, line 79: did we create an issue for this?

ruvi-d: Yes. Linked in the PR description. #237

Kishanthan (approved):

PR #244, opened 2021-04-07
Enable cross language LTO for compiled ballerina apps

### LINKED ISSUE #154: Experiment with function inlining and LTO
**Description:**

As of PR #139, all trivial functions were moved out of the header, which prevents automatic in-lining of these functions. Need to perform a tests that measure compile times and performance with using link time optimization (LTO) vs explicit function in-lining.

**Describe your problem(s)**

**Describe your solution(s)**

### LINKED ISSUE #209: Support LTO between Codegen C and Rust runtime lib
**Description:**

**Describe your problem(s)**

**Describe your solution(s)**

ruvi-d:
## Purpose

This PR enables Link Time Optimization (specifically ThinLTO) between the runtime lib and ballerina apps. The `clang` compiler and the `lld` linker must be used for the this to work (no `gcc` or `ld` support). The clang's LLVM and rustc's LLVM version must also match. I've tested successfully with clang 11.0.0 and rust 1.49.0. 

Other changes: 
* The runtime lib is now built as a static lib (not a dynamic lib)
* When using cmake to build the runtime, the cargo build output is now shown in the std output  

Resolves #209 
Resolves #154

### DIFF
--- a/test/testRunScript.sh
+++ b/test/testRunScript.sh
@@ -21,8 +21,6 @@ else
   fi
 fi
 
-
-
 #Ignoring Ballerina compiler errors due to the use of undefined external functions for print routines
 #if [ -s ./err.log ]
 #then
@@ -39,7 +37,7 @@ then
   exit 1
 fi
 
-clang -O0 -o $filename.out $filename-bir-dump.ll -L../../../runtime/target/release/ -lballerina_rt 2>clang_err.log
+clang-11 --target=x86_64-unknown-linux-gnu -c -O3 -flto=thin -Wno-override-module -o $filename.o $filename-bir-dump.ll 2>clang_err.log
 
 if [ -s ./clang_err.log ]
 then
@@ -48,8 +46,14 @@ then
   exit 1
 fi
 
-export LD_LIBRARY_PATH=../../../runtime/target/release
+clang-11 -flto=thin -fuse-ld=lld-11 -L../../runtime/release/ -lballerina_rt -lpthread -ldl -o $filename.out -O3 $filename.o 2>lld_err.log
+if [ -s ./lld_err.log ]
+then
+  echo "Linker error/warning. Error msg: "
+  cat ./lld_err.log
+  exit 1
+fi
 
 ./$filename.out
 
-rm $filename-bir-dump.ll $filename.out
+rm $filename-bir-dump.ll $filename.out $filename.o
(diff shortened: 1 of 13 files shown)


test/testRunScript.sh
```diff
@@ -39,7 +37,7 @@ then
   exit 1
 fi
 
-clang -O0 -o $filename.out $filename-bir-dump.ll -L../../../runtime/target/release/ -lballerina_rt 2>clang_err.log
+clang-11 --target=x86_64-unknown-linux-gnu -c -O2 -flto=thin -Wno-override-module -o $filename.o $filename-bir-dump.ll 2>clang_err.log
```
ayonam, line 40: You may want to try a -O3 here, which will perform cross-function optimization.  That could give you better code.

test/testRunScript.sh
```diff
@@ -48,8 +46,14 @@ then
   exit 1
 fi
 
-export LD_LIBRARY_PATH=../../../runtime/target/release
+clang-11 -flto=thin -fuse-ld=lld-11 -L../../runtime/release/ -lballerina_rt -lpthread -ldl -o $filename.out -O2 $filename.o 2>lld_err.log
```
ayonam, line 49: Ditto here as the previous comment on -O3

ayonam (approved):

test/testRunScript.sh
ruvi-d: fixed

test/testRunScript.sh
ruvi-d: fixed

ruvi-d:
Looks like we have a bug with map<int> now after #235 was merged to master.
Enabling thinLTO causes the map pointer passed to the runtime (for map load) to be invalid.

ruvi-d:
> Looks like we have a bug with map now after #235 was merged to master.
> Enabling thinLTO causes the map pointer passed to the runtime (for map load) to be invalid.

Resolved the issue by changing all references to `LLVMInt32Type` with `LLVMInt64Type`

shubhamnarlawar:
> > Looks like we have a bug with map now after #235 was merged to master.
> > Enabling thinLTO causes the map pointer passed to the runtime (for map load) to be invalid.
> 
> Resolved the issue by changing all references to `LLVMInt32Type` with `LLVMInt64Type`

Looks fine to me!

Kishanthan (approved):

PR #254, opened 2021-04-19
Use C++ LLVM APIs and Misc changes

### LINKED ISSUE #144: Use C++ LLVM APIs instead of C wrapper API
**Description:**

The code base currently uses a mixture of LLVM C++ and C APIs. This has resulted in the use of a lot of `wrap` and  `unwap` function calls to convert data between the two types of APIs. The C APIs are less type safe. Suggest that we migrate to using only C++ LLVM APIs

**Suggested Labels:**

Type/Improvement

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

ruvi-d:
## Purpose

**Changes:**
* Replace all usages of LLVM C APIs with C++ APIs
* Fix all memory leaks
  * Now it's possible to run checks with sanitizers ON
* Fix bug with linking runtime lib
  * Currently the build script will only look in the cargo's release target folder; even for debug builds
  * With this change the correct runtime target will be linked
* Replace weak_prt with references to hold parent class members
  * Improves performance by removing the extra checks required to lock the weak_ptr
* Reorg classes to reduce the coupling between codegen & BIR classes
* Improve the usability of nBallerina's custom build option flags using CMake's Option feature
* Fixed bug where binary comparison operators where generating unsigned comparison code instead of signed comparisons

Resolves #144


shubhamnarlawar:
> ## Purpose
> 
> **Changes:**
> 
>     * Replace all usages of LLVM C APIs with C++ APIs
> 
>     * Fix all memory leaks
>       
>       * Now it's possible to run checks with sanitizers ON
> 
>     * Fix bug with linking runtime lib
>       
>       * Currently the build script will only look in the cargo's release target folder; even for debug builds
>       * With this change the correct runtime target will be linked
> 
>     * Replace weak_prt with references to hold parent class members
>       
>       * Improves performance by removing the extra checks required to lock the weak_ptr
> 
>     * Reorg classes to reduce the coupling between codegen & BIR classes
> 
>     * Improve the usability of nBallerina's custom build option flags using CMake's Option feature
> 
>     * Fixed bug where binary comparison operators where generating unsigned comparison code instead of signed comparisons
> 
> 
> Resolves #144

If you have fixed all the memory leaks, shall we make sanitizers check default ON in the PR workflow?

So that workflow file checks possible memory leaks in future code. (i.e. simply fails when there are memory leaks)

compiler/codegen/CodeGenUtils.cpp
```diff
@@ -0,0 +1,236 @@
...
+    case TYPE_TAG_NIL:
+        return llvm::Type::getInt8PtrTy(context);
+    case TYPE_TAG_ANY:
+    case TYPE_TAG_UNION:
+    case TYPE_TAG_ANYDATA: {
+        auto *type = module.getTypeByName("struct.smtPtr");
+        if (type != nullptr) {
+            return type;
+        }
+        return llvm::StructType::create(
```
KavinduZoysa, line 60: Don't we need to add the type (struct.smtPtr) here?

ruvi-d:
> If you have fixed all the memory leaks, shall we make sanitizers check default ON in the PR workflow?
> 
> So that workflow file checks possible memory leaks in future code. (i.e. simply fails when there are memory leaks)

Good point. We will have to take it up separately though because we only supporting sanitizers for debug builds. Maybe we can run both release and debug builds for the PR check. I'll create a task to track it.

compiler/codegen/CodeGenUtils.cpp
ruvi-d: Done in line 62. Part of the args of `llvm::StructType::create`

KavinduZoysa: That means, the method `llvm::StructType::create()` will add the type to module and we do not need to call `addTypeName()` right?

ruvi-d: Yep

KavinduZoysa:
@ruvi-d, are you going to push any change to this PR regarding Shubham's suggestion?

ruvi-d:
> @ruvi-d, are you going to push any change to this PR regarding Shubham's suggestion?

No. I'd rather take it separately. This patch is already too large :-)

KavinduZoysa:
> > @ruvi-d, are you going to push any change to this PR regarding Shubham's suggestion?
> 
> No. I'd rather take it separately. This patch is already too large :-)

Okay, have you created an issue?

KavinduZoysa (approved):

ruvi-d:
> Okay, have you created an issue?
Just did. #262

PR #282, opened 2021-04-27
Decouple the codegen logic from BIR classes

### LINKED ISSUE #263: Completely decouple the codegen logic from BIR classes
**Description:**

$subject

**Describe your problem(s)**
Currently some BIR classes have LLVM type member variables. This results in the BIR classes holding codegen state. Thus in the codegen logic to have references to parent BIR classes. This makes it very hard to inject new branching logic when doing codegen.

**Describe your solution(s)**
Use the visitor design pattern to decouple the codegen logic from the BIR classes.

ruvi-d:
## Purpose
Purpose described in #263 

Resolves #263 

## Approach

**Changes**
* Refactor the existing classes following an "acyclic visitor pattern" Ref: [link](https://condor.depaul.edu/dmumaugh/OOT/Design-Principles/acv.pdf)
  * The BIR class hierarchy (visitee classes) now resides in the `compiler/bir` folder, with the header files in `compiler/include/bir`
  * The codegen class hierarchy (visitor classes) now resides in the `compiler/codegen` folder, with the header files in `compiler/include/codegen`
  * The visitor and visitee interfaces are defined in `compiler/include/interfaces/Translatable.h` 
  * Variadic templates were used to automatically generate the visitor/visitee boilerplate logic
  * Double dispatch (visitor pattern) was only use for the instruction classes; with Package, Function and BasicBlock classes using a single dispatch to reduce the overhead
* Codegen for branching Binary operations were simplified since bools are now represented using u1 
* Changed SmartStruct definition to hold the address into the string table as opposed to the numeric offset
  * Since the offset value was already using u64, changing it to hold the address instead, reduced overhead of having to repeatedly codegen the pointer arithmatic logic to generate the said memory address
* Many misc code clean ups


compiler/include/codegen/NonTerminatorInsnCodeGen.h
```diff
@@ -0,0 +1,56 @@
...
+    void visit(class MapStoreInsn &obj, llvm::IRBuilder<> &builder) override;
+    void visit(class MoveInsn &obj, llvm::IRBuilder<> &builder) override;
+    void visit(class StructureInsn &obj, llvm::IRBuilder<> &builder) override;
+    void visit(class TypeCastInsn &obj, llvm::IRBuilder<> &builder) override;
+    void visit(class UnaryOpInsn &obj, llvm::IRBuilder<> &builder) override;
+};
+
+} // namespace nballerina
+
+#endif //!__NONTERMINATORINSNCODEGEN__H__
```
KavinduZoysa, line 56: Please add a new line here

KavinduZoysa (approved):

ramakotareddy (approved):

PR #284, opened 2021-04-27
Inject abort call within the typecast instruction scope

### LINKED ISSUE #265: Investigate possibility to inject branching logic at instruction codegen scope
**Description:**

There is a general need to further "de-sugar" a BIR instruction into branching conditions and additional instruction generations. e.g. calling abort when `is_same_type` check is false and when doing a string value comparison.  

One option is to do it as a BIR pass that de-sugars the instructions into multiple BIR instructions. The implementation for this would be relatively straightforward, but it would introduce a hard to maintain coupling between the BIR pass and the actual instruction codegen logic.

Another option that is currently being used is to do it as a LLVM pass. We are currently using this approach for the `is_same_type` checks. However this approach has limited usefulness for general cases as we would have lost the BIR class/type info required to inject more advance conditional logic.

This task proposes another alternative solution. The proposal is to inject the branching logic at the instruction codegen scope itself. This is not straightforward and may require #263   

**Describe your task(s)**
Suggest that we try to porting the existing `is_same_type` implementation

#263 

@ruvi-d

ruvi-d:
## Purpose
Inject the branching logic at the instruction codegen scope; without having to do a separate pass on the BIR objects or LLVM objects.

Resolves #265 

## Related PRs

Depends on #282


compiler/codegen/FunctionCodeGen.cpp
```diff
@@ -59,11 +59,13 @@ llvm::Value *FunctionCodeGen::getLocalOrGlobalVal(const Operand &op) const {
...
 }
 
+llvm::Function *FunctionCodeGen::getFunctionValue() { return llvmValue; }
+
 void FunctionCodeGen::visit(Function &obj, llvm::IRBuilder<> &builder) {
 
     llvm::Module &module = parentGenerator.getModule();
-    auto *llvmFunction = module.getFunction(obj.name);
-    auto *BbRef = llvm::BasicBlock::Create(module.getContext(), "entry", llvmFunction);
+    llvmValue = module.getFunction(obj.name);
```
KavinduZoysa, line 67: This variable name does not make sense since it getting the function. Shallan we change it?

ruvi-d: @KavinduZoysa  Renamed to `llvmFunction`. Is that ok?

KavinduZoysa: +1.

KavinduZoysa (approved):

KavinduZoysa (approved):

PR #285, opened 2021-04-29
Represent array struct in LLVM side and migrate rust array to match

### LINKED ISSUE #272: Represent Array struct in llvm side
**Description:**

```
typedef DynamicBalArray<T> struct {
    [64bit] header;
    [64bit] BalType *inherentType;
    [64bit] size_t length;
    [64bit] size_t capacity;
    [64bit] struct DynamicArray<T> *array;
};

struct DynamicArray<T> {
   [64bit] unsigned header; 
   [64bit]* T values[0];
};
```

**Describe your problem(s)**

**Describe your solution(s)**

### LINKED ISSUE #278: Migrate rust array to new sturcutre
**Description:**
change the rust side return structure that matches #272

**Describe your problem(s)**

**Describe your solution(s)**

heshanpadmasiri:
## Purpose
Represent the arrays in LLVM structs and make the runtime library create objects matching that type
Resolves #272
Resolves #278

## Approach
* Change the LLVM type definition for array from i8 pointer to a pointer to a struct given in #272
* Change the rust runtime to create, store and load from such a pointer (currently implemented only for int)

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no


runtime/src/bal_array.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+            if self.length <= index {
+
+               self.length = index+1;
+            }
+
+        }
+    }
+
+}
```
ruvi-d, line 108: Missing EoF

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                unsafe{
+                    return *self.array.values.offset(isize::try_from(index).unwrap());
+                }
+            }
+        }
+
+        pub fn set_element(&mut self, index:i64, value:i64){
+            if self.capacity <= index {
+               let new_size = self.capacity * GROWTH_FACTOR;
```
ruvi-d, line 90: Have to keep growing till new index fits

runtime/src/lib.rs
```diff
@@ -109,12 +113,10 @@ pub extern "C" fn print_boolean(num8: i8) {
...
 #[no_mangle]
-pub extern "C" fn array_init_int(size: i64) -> *mut Vec<i64> {
-    let size_t = if size > 0 { size } else { 8 };
-    let size_t = size_t as usize;
-    let foo: Box<Vec<i64>> = Box::new(Vec::with_capacity(size_t));
-    let vec_pointer = Box::into_raw(foo);
-    return vec_pointer as *mut Vec<i64>;
+pub extern "C" fn array_init_int(size: i64) -> *mut DynamicBalArray {
+    let array: Box<DynamicBalArray> = Box::new(DynamicBalArray::new(size));
+    let vec_pointer = Box::into_raw(array);
```
ruvi-d, line 118: vec_pointer is a bad variable name here

runtime/Cargo.toml
```diff
@@ -14,3 +14,4 @@ crate-type = ["staticlib"]
 num = "0.3"
 num-traits = "0.2"
 num-derive = "0.3"
+libc = "0.2.94"
```
ruvi-d, line 17: Missing EoF

compiler/codegen/ArrayInsns.cpp
```diff
@@ -41,7 +41,9 @@ void ArrayInsn::translate(llvm::Module &module, llvm::IRBuilder<> &builder) {
     TypeTag memberTypeTag = lhsVar.getType().getMemberTypeTag();
     auto arrayInitFunc = CodeGenUtils::getArrayInitFunc(module, memberTypeTag);
     auto *newArrayRef = builder.CreateCall(arrayInitFunc, llvm::ArrayRef<llvm::Value *>({sizeOpValueRef}));
-    builder.CreateStore(newArrayRef, lhsOpRef);
+    auto *bitCastRes = builder.CreateBitCast(
+        newArrayRef, llvm::PointerType::getUnqual(module.getTypeByName("struct.dynamicBalArray")), "");
+    builder.CreateStore(bitCastRes, lhsOpRef);
```
ruvi-d, line 46: I assume the bitcase is not required here?

ruvi-d:
One more thing.
Need to support this : https://github.com/ballerina-platform/nballerina/blob/master/test/array/load_store_int_array_print_empty_value.bal

So when you are expanding the array, you need to fill it with empty values too. i.e. 0 for int types

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,108 @@
...
+               DynamicBalArray::copy_memory(self.array.values, new_array, self.capacity);
+               DynamicBalArray::deallocate_memory(self.array.values);
+               self.array.values = new_array;
+               self.capacity = new_size;
+            }
+            unsafe {
+                *self.array.values.offset(isize::try_from(index).unwrap()) = value;
+            }
+            if self.length <= index {
+
```
KavinduZoysa, line 101: This new line is not needed.

runtime/src/lib.rs
heshanpadmasiri: changed to array_pointer

runtime/src/bal_array.rs
heshanpadmasiri: fixed

runtime/src/bal_array.rs
heshanpadmasiri: fixed

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,108 @@
...
+                panic!("Index out of range {} {}", self.length, index);
+            } else {
+                unsafe{
+                    return *self.array.values.offset(isize::try_from(index).unwrap());
+                }
+            }
+        }
+
+        pub fn set_element(&mut self, index:i64, value:i64){
+            if self.capacity <= index {
```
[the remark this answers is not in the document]
heshanpadmasiri: That if block is trying to see if we need to grow the underlying memory block to store the value and I am taking the size of that as the capacity

runtime/Cargo.toml
heshanpadmasiri: fixed

compiler/codegen/CodeGenUtils.cpp
```diff
@@ -61,6 +60,29 @@ llvm::Type *CodeGenUtils::getLLVMTypeOfType(TypeTag typeTag, llvm::Module &modul
             context, llvm::ArrayRef<llvm::Type *>({llvm::Type::getInt64Ty(context), llvm::Type::getInt8PtrTy(context)}),
             "struct.smtPtr");
     }
+    case TYPE_TAG_ARRAY: {
+        auto *dynamicBalArrayType = module.getTypeByName("struct.dynamicBalArray");
+        if (dynamicBalArrayType == nullptr) {
+            auto *dynamicArrayType = module.getTypeByName("struct.dynamicArray");
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

compiler/codegen/CodeGenUtils.cpp
```diff
@@ -61,6 +60,29 @@ llvm::Type *CodeGenUtils::getLLVMTypeOfType(TypeTag typeTag, llvm::Module &modul
             context, llvm::ArrayRef<llvm::Type *>({llvm::Type::getInt64Ty(context), llvm::Type::getInt8PtrTy(context)}),
             "struct.smtPtr");
     }
+    case TYPE_TAG_ARRAY: {
+        auto *dynamicBalArrayType = module.getTypeByName("struct.dynamicBalArray");
+        if (dynamicBalArrayType == nullptr) {
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

compiler/codegen/ArrayInsns.cpp
heshanpadmasiri: Fixed by changing the array init function signature in the cpp side to return a struct pointer type and removing the bit cast

heshanpadmasiri:
> One more thing.
> Need to support this : https://github.com/ballerina-platform/nballerina/blob/master/test/array/load_store_int_array_print_empty_value.bal
> 
> So when you are expanding the array, you need to fill it with empty values too. i.e. 0 for int types

Since I am using calloc to allocated memory it should fill entire block with 0

runtime/src/bal_array.rs
heshanpadmasiri: Fixed

runtime/src/bal_array.rs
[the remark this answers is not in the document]
ruvi-d: got it

compiler/codegen/CodeGenUtils.cpp
```diff
@@ -104,7 +124,7 @@ llvm::FunctionCallee CodeGenUtils::getAbortFunc(llvm::Module &module) {
 llvm::FunctionCallee CodeGenUtils::getArrayInitFunc(llvm::Module &module, TypeTag memberTypeTag) {
     const auto arrayTypeFuncName = "array_init_" + Type::getNameOfType(memberTypeTag);
     auto *funcType =
-        llvm::FunctionType::get(llvm::Type::getInt8PtrTy(module.getContext()),
+        llvm::FunctionType::get(llvm::PointerType::getUnqual(module.getTypeByName("struct.dynamicBalArray")),
```
ruvi-d, line 127: Genrally we want to avoid using magic values like  "struct.dynamicBalArray".
Use `getLLVMTypeOfType(TYPE_TAG_ARRAY) to get the type here. Otherwise if the name changes, you will have to update in two places.

heshanpadmasiri: Fixed

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,107 @@
...
+        header: u64,
+        values: *mut i64,
+    }
+
+    impl DynamicBalArray {
+        pub fn new(size: i64) -> DynamicBalArray {
+            let size_t = if size > 0 { size } else { 8 };
+            let dynamic_array = DynamicArray {
+                header: 0,
+                values: DynamicBalArray::allocate_memory(size_t),
```
ruvi-d, line 45: I was going through the design again and I think we have a problem here. 

I think we need to allocate memory within `DynamicArray` and not allocate memory externally.
As in for a Dynamic array of capacity 4, the structure should look like
```
struct DynamicArray {
        header: u64,
        values1: i64,
        values2: i64,
        values3: i64,
        values4: i64,
    }
```
Otherwise with `values: *mut i64,`, there is no point of having a DynamicArray struct.

@manuranga  your input please

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,107 @@
...
+                    return *self.array.values.offset(isize::try_from(index).unwrap());
+                }
+            }
+        }
+
+        pub fn set_element(&mut self, index: i64, value: i64) {
+            if self.capacity <= index {
+                let repeat = (index / (self.capacity * GROWTH_FACTOR)) + 1;
+                let new_size = self.capacity * GROWTH_FACTOR * repeat;
+                let new_array = DynamicBalArray::allocate_memory(new_size);
```
ruvi-d, line 93: `realloc` should be a better choice here 
https://doc.rust-lang.org/stable/std/alloc/fn.realloc.html

runtime/src/bal_array.rs
ruvi-d: e.g. from the BalMap implementation 
https://github.com/ruvi-d/nballerina/blob/ruvi/task/new_balmap/runtime/c_rt/include/balmap.h#L15
https://github.com/ruvi-d/nballerina/blob/ruvi/task/new_balmap/runtime/c_rt/src/balmap.c#L44

heshanpadmasiri: For clarification is this is just move the memory management logic (memeory allocation and resizing) in to DynamicArray (from DynamicBalArray) where the DynamicArray will still have a pointer to a memory block with all the values (which we then access by pointer offset) or should the  structure of DynamicArray change such that each value in it is a separate variable

ruvi-d: > or should the structure of DynamicArray change such that each value in it is a separate variable

I don't think that is possible unless you use something like C++ templates. Even if you did implement it using templates, you wouldn't be able to call generic functions from LLVM.
So, I think what you need is something like https://en.wikipedia.org/wiki/Flexible_array_member, which is what I've used in the `balmap` example I linked above.

heshanpadmasiri: Changed such that entire DynamicArray is [allocated directly](https://github.com/heshanpadmasiri/nballerina/blob/6110191fe21a49ee1ad021b4cb6719e932488135/runtime/src/bal_array.rs#L45) and [resized directly](https://github.com/heshanpadmasiri/nballerina/blob/6110191fe21a49ee1ad021b4cb6719e932488135/runtime/src/bal_array.rs#L73)

runtime/src/bal_array.rs
heshanpadmasiri: Fixed with [grow_array ](https://github.com/heshanpadmasiri/nballerina/blob/6110191fe21a49ee1ad021b4cb6719e932488135/runtime/src/bal_array.rs#L69) function

compiler/codegen/ArrayInsnsCodeGen.cpp
```diff
@@ -42,9 +42,11 @@ void NonTerminatorInsnCodeGen::visit(ArrayLoadInsn &obj, llvm::IRBuilder<> &buil
 
     auto *lhsOpRef = functionGenerator.getLocalOrGlobalVal(obj.lhsOp);
     auto *rhsOpTempRef = functionGenerator.createTempVal(obj.rhsOp, builder);
+    auto *bitCastRhsOpTempRef = builder.CreateBitCast(
+        rhsOpTempRef, builder.getInt8PtrTy(), "");
```
ruvi-d, line 46: Can drop the third arg since it's the default value: https://llvm.org/doxygen/classllvm_1_1IRBuilderBase.html#a06d64f226fabd94d694d6abd82e3c7b3

compiler/codegen/ArrayInsnsCodeGen.cpp
```diff
@@ -61,9 +63,11 @@ void NonTerminatorInsnCodeGen::visit(ArrayStoreInsn &obj, llvm::IRBuilder<> &bui
     auto *memVal = Type::isSmartStructType(rhsOpTypeTag) ? functionGenerator.getLocalOrGlobalVal(obj.rhsOp)
                                                          : functionGenerator.createTempVal(obj.rhsOp, builder);
     auto *lhsOpTempRef = builder.CreateLoad(lhsOpRef);
+    auto *bitCastLhsOpTempRef = builder.CreateBitCast(
+        lhsOpTempRef, builder.getInt8PtrTy(), "");
```
ruvi-d, line 67: Ditto about ""

ruvi-d (changes requested):

compiler/codegen/ArrayInsnsCodeGen.cpp
heshanpadmasiri: Fixed

compiler/codegen/ArrayInsnsCodeGen.cpp
heshanpadmasiri: Fixed

compiler/codegen/CodeGenUtils.cpp
```diff
@@ -62,6 +61,27 @@ llvm::Type *CodeGenUtils::getLLVMTypeOfType(TypeTag typeTag, llvm::Module &modul
             llvm::ArrayRef<llvm::Type *>({llvm::Type::getInt8PtrTy(context), llvm::Type::getInt8PtrTy(context)}),
             "struct.smtPtr");
     }
+    case TYPE_TAG_ARRAY: {
```
ruvi-d, line 64: Better make a note that as of this PR this new array type is only valid for `int[]` type; Other types like `float[]` would be still using `Int8PtrTy`; even though we are are not handling it explicitly.

compiler/codegen/ArrayInsnsCodeGen.cpp
```diff
@@ -42,9 +42,11 @@ void NonTerminatorInsnCodeGen::visit(ArrayLoadInsn &obj, llvm::IRBuilder<> &buil
 
     auto *lhsOpRef = functionGenerator.getLocalOrGlobalVal(obj.lhsOp);
     auto *rhsOpTempRef = functionGenerator.createTempVal(obj.rhsOp, builder);
+    auto *bitCastRhsOpTempRef = builder.CreateBitCast(
+        rhsOpTempRef, builder.getInt8PtrTy());
```
ruvi-d, line 46: Make a note that this is also only a temp requirement as we can update the load and store function declaration to pass a DynamicBalArray type pointer.

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,108 @@
...
+        inherent_type: BalType,
+        length: i64,   // largest index of an element stored in the array + 1
+        capacity: i64, // amount of memory actually allocated for
+        array: *mut DynamicArray,
+    }
+
+    struct DynamicArray {
+        header: u64,
+        values: [i64],
+    }
```
ruvi-d, line 38: These two structs should have `#[repr(C)]` annotations so that their alignment is consistent with what we defined in LLVM

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+
+        fn get_layout(capacity: i64) -> Layout {
+            let header_layout = Layout::new::<i64>();
+            let align = mem::align_of::<i64>();
+            let size = usize::try_from(capacity).unwrap() * mem::size_of::<i64>();
+            let values_layout = Layout::from_size_align(size, align).unwrap();
+            return header_layout.extend(values_layout).unwrap().0;
+        }
```
ruvi-d, line 67: Need someone to review this get_layout logic

ruvi-d (changes requested):
No comments about the

compiler/codegen/CodeGenUtils.cpp
heshanpadmasiri: Add a comment in the file stating the current state of implementation

compiler/codegen/ArrayInsnsCodeGen.cpp
heshanpadmasiri: Changed the function definition to use the new type and add the comment there as well saying for other array types it is silently ignored

runtime/src/bal_array.rs
heshanpadmasiri: Fixed

ruvi-d (approved):
Need to address : https://github.com/ballerina-platform/nballerina/pull/285/files#r625517875

runtime/src/bal_array.rs
KavinduZoysa: @heshanpadmasiri, can you please explain it here?

heshanpadmasiri: In order to use the rust [alloc](https://doc.rust-lang.org/alloc/alloc/fn.alloc.html) I have to give it a [layout](https://doc.rust-lang.org/std/alloc/struct.Layout.html) of how that memory block should look like. This function gives the memory layout for a [DynamicArray](https://github.com/heshanpadmasiri/nballerina/blob/55e2d7c592285530a4c921dc9ff8c91563e7b37f/runtime/src/bal_array.rs#L37) of a given capacity. 

Since the header is i64 and it's layout is defined in rust i used it directly as the header_layout
For values since it is a i64 array I used the array layout as described [here](https://doc.rust-lang.org/reference/type-layout.html). (Basically it says the alignment is the same as the data type and size is n time the size of type. Since size of returns a usize I thought n must also be usize)

Finally to create the layout for the DynamicArray I extended the layout for header with the layout for value which should give the memory layout for a object with header fallowed by the values

KavinduZoysa: Thanks got it. Can't we set the type of capacity as u64 and do a change like this? 

`let size = capacity * mem::size_of::<i64>();`

heshanpadmasiri: We can't because `mem::size_of::<i64>()` returns a `usize` and it can be multiplied only with another `usize`

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,116 @@
...
+    struct DynamicArray {
+        header: u64,
+        values: [i64],
+    }
+
+    impl DynamicBalArray {
+        pub fn new(size: i64) -> DynamicBalArray {
+            let size_t = if size > 0 { size } else { 8 };
+            let layout = DynamicBalArray::get_layout(size_t);
+            unsafe {
```
manuranga, line 46: ideally `unsafe` should be short as possible. does all of this code needs to be in unsafe.

runtime/src/bal_array.rs
KavinduZoysa: Yes, that's correct. My suggestion was to change the method to accept `u64` instead of i64 `fn get_layout(capacity: u64)`. Then we can simply multiply, right?

runtime/src/bal_array.rs
heshanpadmasiri: The problem is unsafe also defines the scope. I need unsafe to allocate memory in to the `raw_ptr`. And I can't access the `raw_ptr` (which I need to check if the allocation is successful and create the `dynamic_array` pointers) outside of unsafe.

runtime/src/bal_array.rs
heshanpadmasiri: No, what I am saying is if we are to avoid the casting capacity can't be an integer it must be a `usize`. That way it will work fine in the rust side but we may run in to issues if try to access it outside of rust.

compiler/codegen/CodeGenUtils.cpp
```diff
@@ -62,6 +61,33 @@ llvm::Type *CodeGenUtils::getLLVMTypeOfType(TypeTag typeTag, llvm::Module &modul
...
+           Though LLVM ir generated assume all array types to be this type it will be silently ignored by the
+           runtime.
+        */
+        auto *dynamicBalArrayType = module.getTypeByName("struct.dynamicBalArray");
+
+        if (dynamicBalArrayType != nullptr) {
+            return llvm::PointerType::getUnqual(dynamicBalArrayType);
+        }
+
+        assert(module.getTypeByName("struct.dynamicArray") == nullptr);
```
manuranga, line 77: should this be 
```suggestion
        assert(module.getTypeByName("struct.dynamicBalArray") == nullptr);
```

manuranga: Oh i just realized this is the growing part of the array.

runtime/src/bal_array.rs
manuranga: Actually values need not to be aligned, the header is. We have references form stack to header, pointer to value is private.

runtime/src/bal_array.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        pub fn set_element(&mut self, index: i64, value: i64) {
+            if self.capacity <= index {
+                let repeat = (index / (self.capacity * GROWTH_FACTOR)) + 1;
+                let new_size = self.capacity * GROWTH_FACTOR * repeat;
+                self.grow_array(self.capacity, new_size);
+                let old_capacity = self.capacity;
+                self.capacity = new_size;
+                if new_size > old_capacity {
```
manuranga, line 102: better to use something slimier to memcpy, no need to do in this pr, because runtime impl will change. just to comment.

manuranga: oh i got it worng this is remainder, isn't this zeroed out by allocater ?

runtime/src/bal_array.rs
manuranga: btw, out initial idea to allocate header and body in one go might not work, because GC needs to know about them separately.

runtime/src/bal_array.rs
manuranga: ahh I see it's realloc

manuranga (approved):

runtime/src/bal_array.rs
manuranga: following rust code works for me
```
use std::alloc::{alloc_zeroed, Layout};

fn main() {
    let layout = Layout::new::<u16>();
    let raw_ptr;
    unsafe{
        raw_ptr = alloc_zeroed(layout);
    }
    println!("Hello, world! {:#?}", raw_ptr);
}
```
this doesn't need to be fixed right away, because we very likely to rewrite this but just to comment

heshanpadmasiri: Fixed in the byte array implementation

runtime/src/bal_array.rs
heshanpadmasiri: If we are to break the header and body allocation in to two when the array get resized I assume we resize only the value and leave the header intact. Or should we reallocate both again as well.

manuranga: yes we'll only need to reallocate the values. please take a look at map impl's grow logic.
If we reallocate header, we need to update pointers to it from everywhere (ie: like moving GC), which we don't plan to support.

Issue #275, opened 2021-04-23
Keep commonly use llvm sections in a .bc file and inject them to current module


manuranga:
**Description:**
Implement reading a bc file (or even better .ll file) and injecting it to top of current module.
We should be able to call the functions and use always inline to get them used in other functions without performance penalty.

```
unique_ptr<Module> mod = parseIRFile(StringRef("commonFunc.bc" ), error, context);
```

**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

manuranga:
A possible solution : https://lists.llvm.org/pipermail/llvm-dev/2014-July/074993.html

ushirask:
## 1. Start our module on top of the common functions module

```
#include <llvm/Support/SourceMgr.h>
#include <llvm/IRReader/IRReader.h>
...

llvm::SMDiagnostic Err;

auto mod = llvm::parseIRFile("func.ll", Err, *mContext);

mod -> setModuleIdentifier(moduleName);

//Use this mod as the current module used.
```

## 2. Clone function into a prototype
```
#include <llvm/Support/SourceMgr.h>
#include <llvm/IRReader/IRReader.h>
#include <llvm/Transforms/Utils/Cloning.h>
...

llvm::SMDiagnostic Err;
m1Context=std::make_unique<llvm::LLVMContext>();
mod1 = llvm::parseIRFile("func.ll", Err, *m1Context);


 void ReplaceProtoFunc( std::string funcName, llvm::Module* srcModule, llvm::Module* destModule)
{
    llvm::Function *srcFunc = srcModule -> getFunction(funcName);
    auto destFunc = llvm::Function::Create(llvm::cast<llvm::FunctionType>(srcFunc->getType()->getElementType()), srcFunc->getLinkage(), srcFunc->getName(), *destModule);

    llvm::ValueToValueMapTy valuemap;
    llvm::SmallVector<llvm::ReturnInst*, 8> returns;

    llvm::CloneFunctionInto(destFunc, srcFunc , valuemap, true, returns);
    return;
}
```

ruvi-d:
@ushirask Option #2 sounds good. Can you get a PoC up for a function with input params and return value? We can then compare the LLVM IR to make sure that what we are expecting is happening correctly.

[...]

Issue #273, opened 2021-04-23
Migrate String new representation


manuranga:
**Description:**

Migrate string in the on consent section to be represented in the following structure.
Get rid of the `new_string` rust function.
Migrate `print_string` and other rust functions to use this format.

```
typedef BalAsciiString struct {
   [64bit] uint64_t header = 0b000110;
   [64bit] size_t nBytes;
   [nBytes] char bytes[0]; // UTF-8 bytes go here
};
```

**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

shubhamnarlawar:
Reference document - https://docs.google.com/document/d/1wH9T-VTz-Ad9YnrAoT4f05nh6EpBrZxqvfhSzfWng3s/edit?ts=6063f600#heading=h.dyksfb506vbw

shubhamnarlawar:
Process - 

1. Create a pointer to struct `BalAsciiString`.
2. Add `CodeGenUtils::storeValueInBalAsciiString()` to fill members of `BalAsciiString` object.
3. Generate Store instruction
4. Modify rust `print_string` to handle `BalAsciiString` struct object.

Issue #252, opened 2021-04-15
Migrate int-to-any and any-to-int to BalValue representation


manuranga:
**Description:**
Remove fat-pointer (`smtPtr`) usage when compiling int to any casting

**Describe your problem(s)**
Below code should work without any usage of fat pointer and external type checker.

```ballerina
function any_to_int(any a) returns int{
     int i = <int> a;
     return i;
}

function int_to_any(int i) returns any {
     any a = i;
     return a;
}

function test(int i) {
   if (any_to_int(int_to_any(i)) != i) {
        panic error(i.toString());
   }
}

public function main() {
     test(1);
     test(-1);
     test(12147483647);
}
```

**Describe your solution(s)**

Equivalent C++ code to the generated code should look as follows

```C++

unsigned long int_to_any(long i) {
  //  if in immediate mode range :
  //    left shift, or with 1, return
  // else :
  //    allocate on heap and return pointer
}

long any_to_int(unsigned long i) {
   // if ends with 1 :
   //      right shift by 1, return
   // else :
   //     if ends with 00 :
   //          read from heap, return
   //     else:
   //          throw type cast error
}

```
Will add more implementation details instead above pseudo-code as we flesh out this design.

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

ayonam:
Just one small correction.  The left shift and right shift in the pseudo code above needs to interchanged.

ayonam:
Also, the immediate mode range in the pseudo code above can be set to a 61-bit signed integer.  So the range will be   -(2^60)  to  (2^60) - 1.

This is because we expect 64-bit pointers to have their lowest three bits to be zeros.  So we want to use those three bits to denote various things in our scheme of type representation.  Hence, we can represent integers only within the 61-bits that are left.  

In fact, I'd suggest that we do the left/right shifting by 3 bits and use the result as needed.  So replace all the shifts in the pseudo code above with shift by 3 bits.

manuranga:
> Just one small correction. The left shift and right shift in the pseudo code above needs to interchanged.
Ahh yes.

>we can represent integers only within the 61-bits that are left.
I am not seeing the advantage of keeping integers looking the same as the pointers, in fact this can negatively impact by fooling the garbage collector.

manuranga:
I have tested following equivalent C++ code. 
Generated LLVM/x86 can be views form : https://godbolt.org/z/nq6nGMGes
```C++

[...]

PR #288, opened 2021-05-04
Add basic balmap type support via C runtime

### LINKED ISSUE #286: Represent Map struct in llvm side
**Description:**

```C++
// value of null means entry is unused
struct Entry {
  BalString *key;
  BalValue value;
};

// open addressing
struct HashTable {
  // this is gc'ed, so needs GC header
  uint64_t header;
  struct Entry entries[];
};


typedef BalMap struct {
   uint64_t header;   
   // how many entries are used (same as length in Ballerina)
   uint64_t used;
   size_t nEntries;
   struct HashTable *hashTable;
};
```

Ref: [link](https://docs.google.com/document/d/1wH9T-VTz-Ad9YnrAoT4f05nh6EpBrZxqvfhSzfWng3s/edit#heading=h.vspfek914kgd)

**Describe your problem(s)**

**Describe your solution(s)**

ruvi-d:
## Purpose

Replace the existing map<int> support with the new BalMap type as drafted here : https://github.com/ruvi-d/nballerina-colab/tree/main/new_map

Resolves #286 

## Approach

An attempt was made to implement the init, store and load logic via LLVM API calls; but it was pretty evident that this approach would not be easily maintainable. Therefore the logic was implemented in C lang.

Since #275's approach of importing LLVM IR is WIP; this PR explores and alternative approach where we compile the BalMap functions as a static lib and inline the functions at Ballerina app link time. Similar to how we are currently linking the Rust runtime. Thus this is effectively a new "C" runtime.

I was able to verify that LTO is working for this new "C" runtime by replicating the tests from #209 

**Other implementation notes**
* Since Thin LTO support requires clang, the "C" runtime needs to be build with `clang-11` and linked using `lld-11`. Thus to keep the build scripts simple, I've made the change to build nballerina too with clang.
* The Rust runtime was moved from `runtime/` to `runtime/rust_rt/` to allow for the "C" runtime
* gtest unit testing framework was introduced to easily write and verify the "C" runtime via unit tests
  * I think this should be the idiomatic approach to develop and verify runtime functionality; i.e. without having to rely solely on the LIT integration tests.
  * I've setup the tests to run automatically after a build
* I've added placeholder logic for the following missing components
  * BalValue  (awaiting #252 )
  * BalString - compare (awaiting #273 ) 
  * alloc_value and zalloc - (awaiting garbage collector implementation) 
* Map<anydata> type support has been disabled till support for anydata is added via BalValue
* I tried benchmark using this hashtable based implementation vs a simple linear search based implementation
  * I measured the total time taken to insert and find all elements
  * For containers of size smaller than 100, the time measured was less than 1ms for both implementations
  * At the size of 1000 and beyond the hashtable was faster
  * Thus I have removed the linear search logic from this PR


KavinduZoysa (approved):

runtime/c_rt/include/common.h
```diff
@@ -0,0 +1,63 @@
...
+// Static functions
+inline static bool bal_string_equals(BalStringPtr str1, BalStringPtr str2) {
+    return strcmp(str1->value, str2->value) == 0;
+}
+
+// DJB2 hash function
+static unsigned long bal_string_hash(BalStringPtr s) {
+    unsigned long hash = 5381;
+    size_t n = strlen(s->value);
+    const char *p = s->value;
```
KavinduZoysa, line 56:
```
    const char *p = s->value;
    size_t n = strlen(p);
```

Shall we change this?

runtime/c_rt/src/balmap.c
```diff
@@ -0,0 +1,149 @@
...
+static void bal_map_init_hashtable(BalMapPtr map, size_t min_capacity) {
+    map->used = 0;
+    size_t n = HASH_TABLE_MIN_SIZE;
+    while ((size_t)(n * LOAD_FACTOR) < min_capacity) {
+        n <<= 1;
+        if (n == 0) {
+            fprintf(stderr, "Capacity overflow!\n");
+            abort();
+        }
+    }
```
KavinduZoysa, line 36: Can you please explain what you have done here?

runtime/c_rt/src/balmap.c
```diff
@@ -0,0 +1,149 @@
...
+        if (bal_string_equals(key, entries[i].key)) {
+            *outValue = entries[i].value;
+            return true;
+        }
+        if (i > 0) {
+            --i;
+        } else {
+            i = map->n_entries - 1;
+        }
+    }
```
KavinduZoysa, line 125: Shall we assign `map->n_entries` to a local variable?

runtime/c_rt/src/balmap.c
ruvi-d: `n` is the Map capacity that is required to contain `min_capacity` of elements, while maintaining the target load factor. I.e. `(n * LOAD_FACTOR) > min_capacity`. `n` grows in powers of 2. So in line 30-31; we are literately Trying to find the smallest `n` to fit the `min_capacity`.   

Since `n` grows in powers of 2, when it overflows it will be equal to zero. Thus we have the check in line 32-35.

Hope that made sense?

KavinduZoysa: yes, thanks

KavinduZoysa (approved):

PR #290, opened 2021-05-05
Fix broken move semantic support in BIR classes

### LINKED ISSUE #289: Fix broken move semantic support in BIR classes
**Description:**
Throughout the BIRReader class we are creating BIR class objects and calling `std::move` assuming that the transfer will be efficient via move semantics. e.g. https://github.com/ballerina-platform/nballerina/blob/master/compiler/BIRReader.cpp#L556
But since almost all the BIR classes define a default deconstructor, the rule of 3 (https://en.cppreference.com/w/cpp/language/rule_of_three) means that the automatic move semantics would get disabled and that the moves would fallback to copies. 

As such, the BIR classes need to be updated to correctly support move semantics. 

**Steps to reproduce:**

**Affected Versions:**

ruvi-d:
## Purpose
Description #289 

Resolves #289 

## Approach
**Changes**
* Disable copying but allow moving for `AbstractVariable` derived classes and `Function` & `BasicBlock` classes
* `Function` & `BasicBlock` classes are no longer heap allocated to improve performance
* Disable copying and moving for `Package` class since we heap allocate it
* Disable copying and moving for `AbstractInstruction` derived classes since we must heap allocate them anyway to get the double dispatch (observer pattern) working
* Made BIRReader class a friend of `Package`, `Function` and `BasicBlock` classes to simplify BIR class interfaces and to improve their invariance
* Address some warnings thrown by the static analyser for the `BIRReader` class

### DIFF
--- a/compiler/BIRReader.cpp
+++ b/compiler/BIRReader.cpp
@@ -20,6 +20,7 @@
 #include <array>
 #include <cstdio>
 #include <cstdlib>
+#include <queue>
 
 #ifdef unix
 #include <libgen.h>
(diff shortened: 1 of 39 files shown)


compiler/BIRReader.cpp
```diff
@@ -87,27 +81,27 @@ int16_t BIRReader::readS2be() {
 
 // Read 4 bytes from the stream
 int32_t BIRReader::readS4be() {
-    int32_t value;
+    uint32_t value = 0;
```
shubhamnarlawar, line 84: Any reason to change it to unsigned int

ruvi-d: Change it because the static analyzer was complaining about mixing signed and unsigned values. But looks like I've made a mistake here since the return value is supposed to be int32_t. Thanks for catching it.

I've added a new commit overhauling the endianness swapping logic. Please review.

shubhamnarlawar:
The changes looks fine to me.

shubhamnarlawar (approved):

KavinduZoysa (approved):

PR #301, opened 2021-05-10
Extend the array implementation to byte

heshanpadmasiri:
## Purpose
Extend the array implementation in #285 to support byte datatype.
Resolves #295

## Approach
-  Implement the minimum code changes to set a variable as a byte value 
-  Change the array implementation to a generic and use that to implement both integer and byte arrays
-  Add new test cases for byte arrays based on integer arrays

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

## Related PRs
Expand on #285

PR #312, opened 2021-05-17
Implement array access using the C runtime

### LINKED ISSUE #274: Implement array access
**Description:**
Implement array access of int array

Depends on #272

Array access has to depend on `inherentType` if it's not current type it has to cast. This could ether be upcast or a downcast.
**Describe your problem(s)**

**Describe your solution(s)**

heshanpadmasiri:
## Purpose
Allow array access using the type information in the header
Resolves #274 

## Goals
Allow array access when one array is casted to another  using the header of the underlying array

## Approach
* Change the array store functionality to check for actual array type using the header
* Moved the array load function to the C runtime
* Change the array load function to check for both actual array type and destination type

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

## Related PRs
Uses the C runtime from #288
Depend on byte array implementation in #301


test/array/cast_int_array_any.bal.skip
```diff
@@ -0,0 +1,16 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" "%target_variant" "%skip_bir_gen" | filecheck %s
```
ruvi-d, line 1: Why is this test skipped?

test/array/cast_byte_array_int2.bal
```diff
@@ -0,0 +1,14 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" "%target_variant" "%skip_bir_gen" | filecheck %s
```
ruvi-d, line 1: I'd merge `cast_byte_array_int.bal` and `cast_byte_array_int2.bal` together and cover both tests in the one .bal

test/array/cast_byte_array_any.bal.skip
```diff
@@ -0,0 +1,16 @@
+// RUN: "%testRunScript" %s %nballerinacc "%java_path" "%target_variant" "%skip_bir_gen" | filecheck %s
```
ruvi-d, line 1: Ditto about adding skipped test

runtime/rust_rt/src/lib.rs
```diff
@@ -98,14 +98,14 @@ pub extern "C" fn print_boolean(num8: i8) {
 
 #[no_mangle]
 pub extern "C" fn array_init_int(size: i64) -> *mut DynamicBalArray<i64> {
-    let array: Box<DynamicBalArray<i64>> = Box::new(DynamicBalArray::<i64>::new(size));
+    let array: Box<DynamicBalArray<i64>> = Box::new(DynamicBalArray::<i64>::new(size, 3));
```
ruvi-d, line 101: Avoid magic numbers, better use something like a enum for the header value

runtime/rust_rt/src/lib.rs
```diff
@@ -149,31 +149,39 @@ pub extern "C" fn array_init_anydata(size: i32) -> *mut Vec<*mut c_void> {
...
+        mem::forget(arr);
+        let mut new_arr = unsafe { Box::from_raw(new_arr_ptr) };
+        new_arr.set_element(index, i8::try_from(ref_ptr).unwrap());
+        mem::forget(new_arr);
+    }
 }
 
 #[no_mangle]
 pub extern "C" fn array_store_byte(arr_ptr: *mut DynamicBalArray<i8>, index: i64, ref_ptr: i8) {
+    if ref_ptr < 0 {
```
ruvi-d, line 170: ref_ptr needs to be of type `u8 and then we can skip this `< 0`  check

runtime/c_rt/include/gc.h
```diff
@@ -31,6 +31,17 @@
 #define aligned_alloc(a, n) _aligned_malloc(n, a)
 #endif
 
+#define ALLOC_FIXED_VALUE(T) ((T *)alloc_value(sizeof(T)))
+
+void *alloc_array(size_t n_members, size_t member_size) {
```
ruvi-d, line 36: Same as `zalloc`. Rename `zalloc` as `alloc_array`

runtime/c_rt/include/balarray.h
```diff
@@ -0,0 +1,46 @@
...
+ * under the License.
+ */
+#ifndef __BALARRAY__H__
+#define __BALARRAY__H__
+
+#include <stdint.h>
+#include <stdlib.h>
+
+typedef struct DynamicArray {
+    uint64_t header;
```
ruvi-d, line 25: no BalValue elements?

ruvi-d, line 25: Seems that the header for DynamicArray is not required and you can keep the elements pointer in DynamicBalArray itself. Ref: https://github.com/jclark/nballerina-rt/blob/main/balrt.c#L164

runtime/c_rt/src/balarray.c
```diff
@@ -0,0 +1,77 @@
...
+
+void *getItemAt(DynamicBalArray *array_ptr, int64_t index) {
+    if (array_ptr->length <= index) {
+        fprintf(stderr, "%s", "Index is out of range\n");
+        abort();
+    }
+    uint64_t header = array_ptr->header;
+    uint64_t header_type = header & 3;
+    uint64_t *base_ptr = &(array_ptr->array->header) + 1;
+    if (header_type == 0) {
```
ruvi-d, line 38: header values need to go to an enum

runtime/c_rt/include/balarray.h
```diff
@@ -0,0 +1,46 @@
...
+    uint32_t inherent_type;
+    uint64_t length;
+    uint64_t capacity;
+    DynamicArray *array;
+} DynamicBalArray;
+
+void *getItemAt(DynamicBalArray *array_ptr, int64_t index);
+
+int64_t castPointerToValue(DynamicBalArray *array_ptr, void *ptr);
```
ruvi-d, line 38: these two can be static functions in .c. No need to make them external by defining it here.

compiler/reader/ConstantPool.cpp
```diff
@@ -459,15 +459,14 @@ int64_t ConstantPoolSet::getIntCp(int32_t index) {
...
-
 // Search byte from the constant pool based on index
 uint8_t ConstantPoolSet::getByteCp(int32_t index) {
     ConstantPoolEntry *poolEntry = getEntry(index);
     assert(poolEntry->getTag() == ConstantPoolEntry::tagEnum::TAG_ENUM_CP_ENTRY_BYTE);
     auto *byteCp = static_cast<ByteCpInfo *>(poolEntry);
     int32_t value = byteCp->getValue();
-    assert(value>=0 && value<=255);
-    return (uint8_t) value;
+    assert(value >= 0 && value <= 255);
```
ruvi-d, line 468: use the limits macro instead of `0` and `255`: https://en.cppreference.com/w/cpp/types/integer

compiler/bir/Types.cpp
```diff
@@ -135,4 +138,32 @@ void Type::checkMapSupport(TypeTag typeTag) {
     }
 }
 
+bool Type::checkArrayCastSupport(TypeTag destination, TypeTag source) {
```
ruvi-d, line 141: args should be named `srcMemberType` and `destMemberType`

ruvi-d (changes requested):

test/array/cast_int_array_any.bal.skip
heshanpadmasiri: I added these test so when #323 is resolved these too should work

test/array/cast_byte_array_any.bal.skip
heshanpadmasiri: I added these test so when #323 is resolved these too should work

test/array/cast_int_array_any.bal.skip
ruvi-d: isn't this within the scope of this PR? pls correct me if otherwise

runtime/c_rt/include/balarray.h
heshanpadmasiri: Size and alignment of the elements are not known at the c side when compiling. (If it's a int array it's 64bit if a byte array it's 8 bits) We do the memory allocation at the rust side and I use the `getItemAt` function which consider both address of header and content of the header (which tells what is the internal data type) to calculate offsets due to this reason.

heshanpadmasiri: I am using these structs to match the rust side (where we do the actual memory allocation still). So if I drop it here it will cause some changes in the rust side (basically internals of almost all the rust functions will change slightly) so is it okay to do that in this PR given that those changes are not strictly speaking related to array access.

test/array/cast_byte_array_int2.bal
heshanpadmasiri: `cast_byte_array_int.bal` actually testing the array cast instruction working (which currently does a static check to see if types can be casted) for byte to int while `cast_byte_array_int2.bal` don't get a cast instruction so it test whether the array access based purely on the header information is working correctly. So depending on which one fails we can easily figure out if the casting is to blame or not

test/array/cast_int_array_any.bal.skip
heshanpadmasiri: This particular test shouldn't had been skipped it was my bad. The reason for skipping other tests are

1. If the underlying array type is `byte[]` we currently don't have a method to convert byte to any and any to byte
2. If there is a cast of the type `any a = i` not `any[] a = i`  where i is an array because it the case of the former we don't have a method to convert arrays to any. (This is latter where we have an array of any and this is handled)

In these cases I think solving them is out of the scope of this PR

runtime/rust_rt/src/lib.rs
heshanpadmasiri: Fixed

runtime/rust_rt/src/lib.rs
heshanpadmasiri: Fixed

runtime/c_rt/src/balarray.c
heshanpadmasiri: Header type check updated to use enums for the case of bytes (ints are the default case). I couldn't find what are our data types that actually uses 16 and 32 bits so I didn't put enums for them. (We are using `uint64_t header_type = header & 3;` to extract just the last 2 bits because that is the part of the header that tells us the type.)

runtime/c_rt/include/balarray.h
heshanpadmasiri: Fixed

runtime/c_rt/include/gc.h
heshanpadmasiri: This part was removed since it is no longer used

runtime/c_rt/include/balarray.h
ruvi-d: ok

manuranga:
Closeing since project is moving to attic

